Issue #27, opened 2020-06-18
Add docker image security scan to pipeline


grzesuav:
Like :
* https://github.com/aquasecurity/trivy
* https://github.com/goodwithtech/dockle

grzesuav:
https://github.com/Azure/container-scan

grzesuav:
https://github.com/metacontroller/metacontroller/pull/35

grzesuav:
Closing as implemented

PR #35, opened 2020-06-24
chore: #27 - Add docker linters to pipeline

grzesuav:


grzesuav:
Through I think we are good to merge it as it scans the image anyway and results are visible in output

PR #51, opened 2020-11-20
chore(api): Update CRD api versions to v1

fpetkovski:
This PR uses [controller-gen](https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/controller-gen) in conjunction with [kubebuilder annotations](https://book.kubebuilder.io/reference/markers/crd.html) to update CRD versions to `apiextensions.k8s.io/v1`.

The CRDs can be generated using: `controller-gen crd paths=./apis/metacontroller/... output:stdout > crds.yaml`

### DIFF
--- a/hack/get-kube-binaries.sh
+++ b/hack/get-kube-binaries.sh
@@ -10,7 +10,7 @@ set -u
 # The integration test framework expects these binaries to be found in the PATH.
 
 # This is the kube-apiserver version to test against.
-KUBE_VERSION="${KUBE_VERSION:-v1.11.3}"
+KUBE_VERSION="${KUBE_VERSION:-v1.18.8}"
 KUBERNETES_RELEASE_URL="${KUBERNETES_RELEASE_URL:-https://dl.k8s.io}"
 
 # This should be the etcd version downloaded by kubernetes/hack/lib/etcd.sh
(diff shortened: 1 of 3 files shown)


manifests/metacontroller.yaml
```diff
@@ -1,47 +1,517 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: compositecontrollers.metacontroller.k8s.io
+  annotations:
+    "api-approved.kubernetes.io": "unapproved, request not yet submitted"
```
fpetkovski, line 6: Turns out the `k8s.io` namespace has been reserved for official Kubernetes resources.

More information [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190612-crd-group-protection.md#what-to-do-if-you-accidentally-put-an-unapproved-api-in-a-protected-group)

hack/get-kube-binaries.sh
```diff
@@ -10,7 +10,7 @@ set -u
 # The integration test framework expects these binaries to be found in the PATH.
 
 # This is the kube-apiserver version to test against.
-KUBE_VERSION="${KUBE_VERSION:-v1.11.3}"
+KUBE_VERSION="${KUBE_VERSION:-v1.18.8}"
```
fpetkovski, line 13: The api version `apiextensions.k8s.io/v1` is only available from 1.16

PR #56, opened 2020-11-23
chore: Add additional golang linters

grzesuav:


grzesuav:
Some of linters are commented out as they are require fixes to code to make them pass. It can be done one by one later.

AmitKumarDas (approved):

PR #76, opened 2020-11-30
fix(manifests): Fix image tag to match what exists

benley:
metacontrollerio/metacontroller:0.4.5 does not exist on hub.docker.com.  All the tags currently there have a `v` prefix.


grzesuav:
hi @benley , thanks for PR but already have fix here - https://github.com/metacontroller/metacontroller/pull/75

benley:
Ah cool, I'll close this one.

Issue #83, opened 2020-12-04
Unable to set multiple ports in service-per-pod


tobias-zeptio:
When i set multiple ports in the `service-per-pod-ports` annotations, the service creation fails because of missing port names. Are multiple ports supported, and is there a way to add the port names to the annotation? The hook only adds port and targetPort.

fpetkovski:
This is fixed in https://github.com/metacontroller/metacontroller/pull/84

tobias-zeptio:
Great @fpetkovski ! I see it's not being released now, do you know when it will be?

grzesuav:
@tobias-zeptio this is `example` , not production code, nothing to release here

tobias-zeptio:
Ah, of course. My bad!

PR #84, opened 2020-12-04
chore(examples): [#83] - Fix service-per-port sync hook

fpetkovski:
In Later versions of kubernetes services with multiple ports need to have names for their ports.

This MR fixes the service-per-pod example


AmitKumarDas (approved):

grzesuav:
hi @fpetkovski , can you change commit prefix to `chore` as this should not trigger a release ?

grzesuav (changes requested):
change commit prefix to `chore` instead of `fix

fpetkovski:
Sure thing, should be done

grzesuav (approved):

grzesuav:
thanks for contribution !

PR #177, opened 2021-02-04
fix: Add build information

grzesuav:
# to verify
* `docker pull grzesuav/metacontroller-test:version`
* `docker run -it grzesuav/metacontroller-test:version`
*there will be :
```
I0204 20:22:55.546557       1 main.go:58] "Metacontroller build information" version="version"
```

# this PR also :
* remove obsolete vendor usage (with go modules this is not needed)
* use `make install` in stead `go install` (for code reusability)
* update alpine to 3.13.1


AmitKumarDas (approved):
Nice :clap:

PR #229, opened 2021-04-06
chore: Add logcheck to check if logging is in stuctured format

grzesuav:


grzesuav:
changes are related to CI checks, therefore merging

PR #231, opened 2021-04-08
docs: Fix invalid URLs in documentation

fpetkovski:
The documentation found at https://metacontroller.app/ contains raw variables which are not properly interpolated with actual values. Examples can be found in the [installation](https://metacontroller.github.io/metacontroller/guide/install.html#install-metacontroller) and [build](https://metacontroller.github.io/metacontroller/contrib/build.html#building) sections.

Looking at the capabilities of mdbook, there does not seem to be a well supported plugin for working with variables. I therefore decided to simply hardcode the actual values directly in the documentation source.

### DIFF
--- a/docs/src/contrib/build.md
+++ b/docs/src/contrib/build.md
@@ -9,7 +9,7 @@ First, check out the code:
 # place the repo according to the Go import path:
 #   $GOPATH/src/metacontroller.io
 cd $GOPATH/src
-git clone {{ site.repo_url }}.git metacontroller.io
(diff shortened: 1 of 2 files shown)


grzesuav:
thanks @fpetkovski , can you change commit message to `docs :` prefix, as `fix` will cause new version of metacontroller, which isn't a casehere?

fpetkovski:
Sure thing, done 👍

grzesuav (approved):

PR #242, opened 2021-04-20
fix(ControllerRevision): #144 - Fix ControllerRevision management

grzesuav:
* moved `OwnerReference` creation (to be present when new `ControllerRevision` is created)
* added resource version handling
* migrated deprecated compare method


controller/composite/controller.go
```diff
@@ -562,10 +564,14 @@ func (pc *parentController) syncParentObject(parent *unstructured.Unstructured)
 	return manageErr
 }
 
+func (pc *parentController) isUsingGeneratedLabelSelector() bool {
```
grzesuav, line 567: not sure about the name, but could not find any better, looking for inspiration

PR #243, opened 2021-04-21
chore(linters): Change event used to trigger linter actions

grzesuav:
Previous one does not allow to comment on PR's from actions


grzesuav:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

PR #260, opened 2021-05-13
perf(k8sClient): use a cache-based version of k8s client

fpetkovski:
The current metacontrollers use a plain controller-runtime client when talking to the API server.
As a result, each read and list action goes directly to the server instead of using a cache.

This commmit initializes the client to be the same as the one from the controller manager.
In this way, reading actions will always be backed by a cache, while write actions
will be executed directly against the API server.

Signed-off-by: fpetkovski <filip.petkovsky@gmail.com>

### DIFF
--- a/pkg/controller/common/common.go
+++ b/pkg/controller/common/common.go
@@ -92,14 +92,7 @@ func NewControllerContext(configuration options.Configuration, mcClient *mcclien
 	}
 	recorder := broadcaster.NewRecorder(scheme, corev1.EventSource{Component: "metacontroller"})
 
-	// Create a new Kubernetes client for interacting with the Kubernetes API
-	k8sClient, err := client.New(configuration.RestConfig, client.Options{})
-	if err != nil {
-		return nil, err
-	}
-
 	return &ControllerContext{
-		K8sClient:         k8sClient,
 		Resources:         resources,
 		DynClient:         dynClient,
 		DynInformers:      dynInformers,
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -66,6 +66,11 @@ func New(configuration options.Configuration) (controllerruntime.Manager, func()
 		return nil, nil, err
 	}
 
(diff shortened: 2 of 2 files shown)


grzesuav:
hi @fpetkovski , thank you for contribution, I missed the notification email about this one.

The one thing I have doubts here is redundant assignment in context creation method - https://github.com/metacontroller/metacontroller/blob/3fef5b4cabeb158a248bbd4ce7effe5830eef006/pkg/controller/common/common.go#L96

I was thinking how to handle it better, for sure mentioned code should be removed, I just thinking what is the best way to provide client there :
* assign as it is currently
* make `k8sClient` a parameter in `Start` method and pass it there - https://github.com/metacontroller/metacontroller/blob/3fef5b4cabeb158a248bbd4ce7effe5830eef006/pkg/server/server.go#L100



what do you think ?

grzesuav (changes requested):
As in comment above.

fpetkovski:
Hi @grzesuav, yes, that is a good point. I also did not see any good way to do this because there seems to be a circular dependency between the manager (due to broadcaster) and the execution context (due to the client from manager). 

I would lean towards simply removing https://github.com/metacontroller/metacontroller/blob/3fef5b4cabeb158a248bbd4ce7effe5830eef006/pkg/controller/common/common.go#L96

My concern with passing the client in https://github.com/metacontroller/metacontroller/blob/3fef5b4cabeb158a248bbd4ce7effe5830eef006/pkg/server/server.go#L100 is that the entire initialization logic starts to get a bit convoluted and hard to follow.

Wdyt?

grzesuav:
@fpetkovski agreed :+1:

grzesuav (approved):

grzesuav:
thanks @fpetkovski ! Sorry for delay, got busy weekend

fpetkovski:
@grzesuav no worries, was a long weekend for me as well :)

PR #265, opened 2021-05-25
fix(deps): Update k8s and controller-runtime dependencies

grzesuav:


main.go
```diff
@@ -108,7 +108,7 @@ func main() {
 	// before we exit
 	var wg sync.WaitGroup
 	wg.Add(1)
-	mgrStopChan := make(chan struct{})
+	mgrStopChan := signals.SetupSignalHandler()
```
fpetkovski, line 111: :100:

pkg/controller/composite/metacontroller.go
```diff
@@ -85,12 +85,12 @@ func NewMetacontroller(controllerContext common.ControllerContext, mcClient mccl
 	return mc
 }
 
-func (mc *Metacontroller) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (mc *Metacontroller) Reconcile(context context.Context, request reconcile.Request) (reconcile.Result, error) {
```
fpetkovski, line 88: Is the variable here shadowing the `context` package name? Usually the variable is called `ctx` to avoid having the same name as the package.

grzesuav: done

PR #272, opened 2021-05-28
fix(security): Update golang.org/x/crypto because of CVE-2020-29652

fpetkovski:
Fixes the docker lint CI job by addressing CVE-2020-29652


grzesuav:
thanks @fpetkovski , already updated and merged other one , very appreciate your investigation !

PR #291, opened 2021-06-22
fix(metrics): Utilize controller-runtime  metrics server

grzesuav:

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,10 +17,7 @@ limitations under the License.
 package main
 
 import (
-	"context"
-	"errors"
 	"flag"
-	"net/http"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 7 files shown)


fpetkovski:
Are we keeping the existing metrics with this change?

grzesuav:
> Are we keeping the existing metrics with this change?

Even better, we are adding them. Currently very few metrics were exposed, as we use the same k8s client for controller-runtime and metacontroller, and controller-runtime registeres it right, we got all worqueue metric in place ! I will add some example metrics before and after the change

main.go
```diff
@@ -120,26 +114,7 @@ func main() {
...
-		}
-		klog.InfoS("Stopped metrics server")
-	}()
-
-	mgrStopChan.Done()
-	if err = srv.Shutdown(context.Background()); err != nil {
-		klog.ErrorS(err, "Error shutting down...")
-	}
+	<-mgrStopChan.Done()
+	klog.InfoS("Stopped metrics server")
```
fpetkovski, line 118: Should be updated to something like Stopped controller manager or similar

fpetkovski:
Ah that's great! Just one nit, otherwise lgtm

grzesuav:
[new.txt](https://github.com/metacontroller/metacontroller/files/6703316/new.txt)
[old.txt](https://github.com/metacontroller/metacontroller/files/6703317/old.txt)

I ran `examples/configmappropagation` test and collected metrics afterards.

PR #292, opened 2021-06-22
chore(examples): #68 - Add subresources status to example CRD

2ZZ:
The CRD example from the guide is missing the subresources status entry and without it the controller won't work, error is below.
```
metacontroller.go:126] "Ignoring CompositeController" name="hello-controller" reason="subresource 'Status' not enabled" groupVersionKind="example.com/v1, Kind=HelloWorld"
```
Ref https://stackoverflow.com/questions/56427423/metacontroller-how-to-stop-calls-to-sync-hooks-and-resource-generation/

### DIFF
--- a/docs/src/guide/create.md
+++ b/docs/src/guide/create.md
@@ -50,6 +50,8 @@ spec:
     plural: helloworlds
     singular: helloworld
   scope: Namespaced
+  subresources:
+    status: {}
 ```
 
 Then apply it to your cluster:


grzesuav:
hi @2ZZ , thanks for contribution, could you please:
* update commit message and amend commit, just as you did with PR name
* sign-off commit (uf you click on DCO check it will give you instructions)

thanks

2ZZ:
Hi @grzesuav, is it ok now?

grzesuav (approved):

Issue #42, opened 2020-08-31
Helm chart


oridool:
Hello,
Where can I find helm chart for this project? 
The only one I found is a closed PR in helm/charts : https://github.com/helm/charts/pull/8591/files 
But it doesn't seem to be in their master branch .

Thank you.

grzesuav:
hi @oridool , we do not have one yet

sathieu:
@oridool Please test this PR: https://github.com/metacontroller/metacontroller/pull/299

PR #299, opened 2021-07-01
chore: #42 - Add helm chart

sathieu:
Add basic helm chart.

Fixes: #42


grzesuav:
hi @sathieu , thanks for PR, could you use instructions to docs about the chelm chart and how to use it, i.e. which values are customizable and what do they mean ?

helm/metacontroller/crds/metacontroller-crds-v1.yaml
```diff
@@ -0,0 +1,534 @@
+
```
grzesuav, line 1: can't it include this file from `manifests/` directory ?

grzesuav:
also, could you change commit type to `chore`/`bugfix` as it does not bring any code change ?

grzesuav:
also, please rebase on latest master to fix pipelines, just merged fix

fpetkovski:
Should we keep all deployment manifests in the `deploy` folder?
So the tree would look like
```
deploy
-- kustomize
-- helm
```

helm/metacontroller/crds/metacontroller-crds-v1.yaml
sathieu: Unfortunately not (or at least I've not found a way).

sathieu:
@grzesuav I've rebased to fix the CI, renamed `chore`, moved to `deploy/helm` and added docs. Back to you 🏸 !

deploy/helm/metacontroller/templates/statefulset.yaml
```diff
@@ -0,0 +1,55 @@
...
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+          - /usr/bin/metacontroller
+          args:
+          - --logtostderr
```
grzesuav, line 39: Those args should be configurable and allow to set any of the metacontroller args. I recommend to do some templating there and define default set in `values`

grzesuav (approved):
I think that for first version it it good enough, I have just one comment with providing more configurable args options to chart to make it more usable

grzesuav:
hi @sathieu , thanks for this, I have several improvements in mind, but do not want to keep it for too long. They can be addressed as separate PR's.

grzesuav:
> Should we keep all deployment manifests in the `deploy` folder?
> So the tree would look like
> 
> ```
> deploy
> -- kustomize
> -- helm
> ```

good idea, will adjust manifests in next PR

grzesuav:
Created https://github.com/metacontroller/metacontroller/issues/302 for improvements

PR #304, opened 2021-07-10
feat(standalone-mode): Enable using composite controller as a library

fpetkovski:
Currently it is only possible to run Metacontroller as a cluster-wide singleton. 
This mode requires deploying two additional CRDs, `CompositeController` and `DecoratorController`.

This commit makes the changes needed to use composite controller
as a library from third-party applications. It introduces a standalone
mode and adds an example of how it can be used.

The following changes were made to the project structure:
* Expose the `composite.NewParentController` function as public
* Introduce `server/standalone.go` to help bootstrap a standalone controller
* Introduce examples-standalone to illustrate library use. We could merge this into the examples folder as 
```
- examples
-- server
-- standalone
```
but that would require changing lots of references in the docs.

My initial suggestion is to announce this as an alpha feature since it is the first iteration and we don't yet know what could go wrong. Users who are in need of this solution can test it and provide feedback so that we can iron out the rough edges.

Relates to:
* https://github.com/metacontroller/metacontroller/issues/4
* https://github.com/metacontroller/metacontroller/issues/218
* https://github.com/metacontroller/metacontroller/issues/266


grzesuav:
hi @fpetkovski , thanks for contribution !

I will try to look next week or two, as I have busy weekend and want to finish logging change (which will be breaking ones, as they do change starting flags and  I want to clean up deprecated flags in the same release)

fpetkovski:
Hey @grzesuav no worries, take your time. There is no rush for reviewing this change :)

grzesuav:
hi @fpetkovski , can you rebase upon 2.0.0 ? I will take a look in next few days on this

fpetkovski:
@grzesuav the PR should be rebased with all tests passing :)

grzesuav:
I wonder if we could have `library` subdirectory in `examples`, as it would keep it on one place. Otherwise I do not have much suggestions. Want just to node that this will be considered as alpha feature, as I cannot guarantee API stability for now. I have in mind also using `goreleaser`  for building and pushing images, and pushing the docs to `go pkg` site, would also help with using as library, but it will be a rather long-term goal.

grzesuav (changes requested):
Can you move `examples-standalone` into `examples` directory ? Name like `standalone` or `library` would be good enough I think

pkg/dynamic/discovery/discovery.go
```diff
@@ -198,6 +198,16 @@ func (rm *ResourceMap) HasSynced() bool {
 	return rm.groupVersions != nil
 }
 
+func (rm *ResourceMap) WaitForSync() {
```
grzesuav, line 201: I know if it mostly not present in the code, but if we wish to make it an API I would ask to add documentation there in form in godoc comments on public methods

grzesuav:
cold you check if writing a integration test (like in `test/integration`) would work for this library mode ? It would let us know about API changes

grzesuav:
@fpetkovski as this PR is a bit stale I created issue for that https://github.com/metacontroller/metacontroller/issues/604, closing this as it got diverged

fpetkovski:
No problem. This somehow dropped off my radar completely. I can revisit it if a need arises again :)

Issue #255, opened 2021-04-28
Feature proposal: All way to authenticate calls to custom controllers (sync/finalize/customize hooks)


grzesuav:
# Request description

Currently metacontroller uses `http` communication without any authorization to execute hookhs.

# Proposal

Add ways to authenticate, like basic auth, ca certs, bearer token. 

Hook API will change, so need to carefully introduce this change to not make any breaking changes.

MartinKanters:
Thanks for creating this. I agree that all of these means of authentication would be valuable, but these can be fixed in multiple PRs/iterations, right? Otherwise this has the risk of blowing up into a lot of work.

p.s. there's a typo in the title. All -> Add.

grzesuav:
> Thanks for creating this. I agree that all of these means of authentication would be valuable, but these can be fixed in multiple PRs/iterations, right? Otherwise this has the risk of blowing up into a lot of work.

it depends, I want first check how it can be implemented, potentially there will be an easy way to do it in one go, and if not - at least API need to be defined/proposed in first PR, even if it won't be implemented to reduce number of changes in CRD's definition. Of course, implementation of methods can be separated if needed.

> 
> p.s. there's a typo in the title. All -> Add.

thanks :)

MartinKanters:
> it depends, I want first check how it can be implemented, potentially there will be an easy way to do it in one go, and if not - at least API need to be defined/proposed in first PR, even if it won't be implemented to reduce number of changes in CRD's definition. Of course, implementation of methods can be separated if needed.

Makes sense! Thanks, looking forward to it.

grzesuav:
I am thinking of having sth similar as in https://github.com/prometheus/common/blob/main/config/http_config.go#L159 - @MartinKanters do you think it cover most common use cases ?

grzesuav:
Or similar https://github.com/prometheus-operator/prometheus-operator/blob/master/pkg/apis/monitoring/v1/types.go#L665-L675

MartinKanters:
Took a quick look at it and it looks fine to me!

dkhacharia:
I am also waiting for this feature to be able to adopt metacontroller. Is it in the making?

grzesuav:
not yet, focusing on other changes currently. Also, currently touching the logic around the hooks.

NiklasRosenstein:
Is the project still open to a contribution on this?

Maybe we can add a secret ref to the `webhook:` field?

grzesuav:
[...]

PR #297, opened 2021-06-30
feat(logging): #233 - Allow logging in json format

grzesuav:
Signed-off-by: grzesuav <grzesuav@gmail.com>

PR #308, opened 2021-07-15
perf(webhooks): #255 - Create httpClient per controller instead ad-hoc creation

grzesuav:
Up to now, for each invocation new httpClient was created. It was both
unefficient and forbids from further changes (like adding authentication
etc). This PR introduces Hooks Executor which are partof the controller.


grzesuav:
it will be rebased after https://github.com/metacontroller/metacontroller/pull/297 will be merged

Issue #298, opened 2021-07-01
Allow to update status of related objects


sathieu:
This is somewhat a follower of #236.

I have a  `CompositeController` which:

- has parent `CertificateGenerator` resource (which describes in which namespace certs should be generated, and selector for Ingresses)
- has related objects as `Ingress`
- has child object would be `Certificate`

**Feature request**: I want to update the ingress `status`  (to fill `status.loadBalancer.ingress: [ip: '10.20.30.40']`).

[Alternative](https://github.com/metacontroller/metacontroller/issues/236#issuecomment-821275579):

> What about adding a boolean like namespaceAsOwner to the operator spec and if set to true, set the owner reference the the namespace owning the parent object?

sathieu:
Note: This `status` is needed for [ArgoCD to work correctly](https://argoproj.github.io/argo-cd/faq/#why-is-my-application-stuck-in-progressing-state).

grzesuav:
hi @sathieu , as I understand `CertificateGenerator` are cluster scope ? 

The issue I see here metcontroller does not own related objects. How those `Ingreses` are created ?

grzesuav:
I think none of the existing primitive in metacontroller (`Decorator`/`Composite` controllers) and in design one (https://metacontroller.github.io/metacontroller/design/map-controller.html) would fit for this use case tbh.

sathieu:
(NB: I've updated the description to my real use-case: Ingress to Istio)

Hi @grzesuav,

>  How those Ingreses are created ?

Those are mostly created by Helm charts from FLOSS projects.

> I think none of the existing primitive in metacontroller [...] and in design one [...] would fit for this use case tbh.

I thought about a workaround: Use to composite controllers. One having the `Ingress2IstioGenerator` as parent, `Ingress`es as related and `Certificate`s as children. Another `Istio2IngressStatusGenerator` having `Ingress` as parent, nothing as children and `Gateway`s as related. This second composite controller would set the ingress status based on the related Istio gateway. WDYT?

grzesuav:
The thing is, `CompositeController` is suppose to work with user defined CRD's, not the ones handled by decorator controller. I am usnure how it would work when configured like that, I will try to ask on `operators` channel on kubernetes slack what would be good fit for this situation

sathieu:
@grzesuav Digging deeper, my usecase is as follow:

[...]

PR #311, opened 2021-07-21
feat: #236 - Use parent's namespace when children is in another namespace

### LINKED ISSUE #236: Allow to create resources in another namespace
Context: I want to fix https://github.com/jetstack/cert-manager/issues/2522 (or https://github.com/istio/istio/issues/14598): From an ingress in any namespace, the controller should create a certificate in the specified namespace.

Currently this doesn't work, as `ownerReferences` don't work accross namespaces (see https://github.com/kubernetes/kubernetes/issues/100577#issuecomment-808756161) and for good reasons.

My code ensure the created resources don't overlap by using a prefix and by including the namespace in the target name.

Possible solution: Allow to skip attaching an owner to the created resource (store this in an annotation instead). The garbage collection would happen with the labels.

sathieu:
Trying to fix #236, related to #298.

### DIFF
--- a/pkg/controller/common/manage_children.go
+++ b/pkg/controller/common/manage_children.go
@@ -275,8 +275,17 @@ func updateChildren(client *dynamicclientset.ResourceClient, updateStrategy Chil
 				continue
 			}
 
-			// We always claim everything we create.
-			controllerRef := MakeControllerRef(parent)
+			// When the children is in another namespace (and the parent is namespaced),
+			// the real parent is the parent's namespace
+			realParent := parent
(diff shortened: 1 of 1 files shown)


grzesuav:
hi @sathieu , this PR does not seems right to be merged, is is a hack, which does not align with any of the patterns which metacontroller//kubernetes follows. When `parent`  CR is deleted, then children  is not cleaned, I stated earlier (https://github.com/metacontroller/metacontroller/issues/236#issuecomment-821284197) that I do not consider this as valid solution.

PR #321, opened 2021-07-27
fix: Delete metacontroller-crds-v1beta1.yaml

sathieu:
Only deploy v1 CRDs with Helm, this fixes the following ArgoCD error:

`Resource apiextensions.k8s.io/CustomResourceDefinition//compositecontrollers.metacontroller.k8s.io appeared 2 times among application resources.`

(Same for `controllerrevisions `and `decoratorcontrollers`).


grzesuav:
hi @sathieu , thanks for the MR. What do you think of making it optional, like by default deploy `v1`, but if user wants deploy in version `v1beta` ? I think it would be useful for people which still are deploying to older kubernetes clusters

sathieu:
@grzesuav I'm OK with this, but I won't have the time to work on it soon (vacations are very near 😎 ).

grzesuav (approved):

grzesuav:
hi, as more people tend to report this, I will merge it, and if someone would need helm chart for older k8s version I will add parameterisation to choose CRD version

Issue #345, opened 2021-08-27
Allow cluster-scoped related resources for namespaced parents


sathieu:
Given an `Ingress` as parent (namespaced), I want to get access to it's parent `IngressClass` (cluster-scoped).

Some context [here](https://gitlab.com/kubitus-project/kubitus-ingress-operator/-/blob/902ca26cbf045d0367f6239c1461786f8746fc0a/kubitus_ingress_operator/ingress2istio.py#L44-52) and [here](https://gitlab.com/kubitus-project/kubitus-ingress-operator/-/blob/902ca26cbf045d0367f6239c1461786f8746fc0a/kubitus_ingress_operator/ingress2istio.py#L90-107).

grzesuav:
@sathieu I will try to determine if it would be possible. If children will be still namespaced-scope in the same namespace I think it would be doable

Ragatzino:
I'm trying this, it's my first pr : https://github.com/metacontroller/metacontroller/pull/362

grzesuav:
hi @sathieu , I checked the code, how your `resource rule` looks like ? From the code, if parent resource is namespaced, it still can have cluster-wide related resources. The only limitation is, if related resource is namespaced and from another namespace, it does not work currently.

grzesuav:
ah, wait, there is second check in code, need to test it.

grzesuav:
https://github.com/metacontroller/metacontroller/pull/363 - @sathieu @Ragatzino after reviewing code I found a bit simpler way, would appreciate feedback. I will add more unit tests, and documentation

grzesuav:
just realized it will be slightly more complicated, as it potential requires API changes, need to review this case

Ragatzino:
Thanks for the reply :). Tell us when you need a review or smt.

grzesuav:
there is a related discussion in https://github.com/metacontroller/metacontroller/discussions/370

sathieu:
@grzesuav Couldn't proposed PR be changed to be backward compatible?

i.e. only specifying namespace when different from parent namespace  :
```json
{
  "related": {
    "Pod.v1": {
        "resourceInTheParentNamespace": {},
        "someOtherNamespace/someResource": {}
    }
  }
}
```

grzesuav:
@sathieu and how you denote cluster-scoped resource here ?

sathieu:
@grzesuav As they work today, i.e without prefix:

```json
{
  "related": {
[...]

PR #362, opened 2021-09-14
feat: #345 - Allow cluster scoped related resources for namespace…

Ragatzino:
Adding wildcards and no names to be added freely inside manager.go inside matchesRelatedRule(parent,related,relatedRessource) function


grzesuav:
hi @Ragatzino , thanks for the PR, will look around weekend. For sure  missing part is a documentation (https://github.com/metacontroller/metacontroller/blob/master/docs/src/api/customize.md need updating )

grzesuav:
Closing as it would cause an API break atm

PR #363, opened 2021-09-15
feat: #345 - Allow related resources from different namespace and cluster scoped

grzesuav:


grzesuav:
Closing as it would requie an API break in current implementation

Issue #414, opened 2022-01-02
Customize hook response cached with only name and clashes between different namespaces


srfrnk:
**EDIT**: Leaving the full conversation intact but after digging deeper the real issue can be found [here](https://github.com/metacontroller/metacontroller/issues/414#issuecomment-1014242000)

**Previous description:**
I'm getting a strange issue when trying to use `relatedResources` with two different namespaces.

I'm deploying these CRDs:
```yaml
apiVersion: jabos.io/v1
kind: GitRepository
metadata:
  name: first-repo
  namespace: stg
spec:
  url: https://github.com/srfrnk/jabos-examples.git
  branch: stg
---
apiVersion: jabos.io/v1
kind: JsonnetManifest
metadata:
  name: simple-jsonnet-manifest
  namespace: stg
spec:
  gitRepository: first-repo
  path: simple-runtime/manifests
  targetNamespace: stg
  commitTLAKey: latestCommitId
---
apiVersion: jabos.io/v1
kind: GitRepository
metadata:
  name: first-repo
  namespace: example-env
spec:
  url: https://github.com/srfrnk/jabos-examples.git
  branch: main
---
apiVersion: jabos.io/v1
kind: JsonnetManifest
metadata:
  name: simple-jsonnet-manifest
  namespace: example-env
spec:
  gitRepository: first-repo
  path: simple-runtime/manifests
  targetNamespace: example-env
  commitTLAKey: latestCommitId
---
```

`JsonnetManifest` is controlled by a `DecoratorController` whose `customize` hook returns:
```json
{"relatedResources":[{"apiVersion":"jabos.io/v1","resource":"git-repositories","namespace":"stg","names":["first-repo"]}]}
```
and
```json
{"relatedResources":[{"apiVersion":"jabos.io/v1","resource":"git-repositories","namespace":"example-env","names":["first-repo"]}]}
```
Respectively.

However the second CRD deployed (always the second) will show the following `Event` and fail to sync:
```bash
Warning  SyncError  2s (x11 over 7s)  metacontroller  Sync error: requested related object namespace stg differs from parent object namespace example-env
```

What's wrong here? Should the related object namespace not be just the parent namespace with this configuration?

srfrnk:
Furthermore - after that happens when deleting the resource the delete will stall...
Logs seems to indicate that the `finalize` hook is never called for that resource.

srfrnk:
Interestingly enough if I change the response from the `customize` hook to:
```json
[...]

PR #419, opened 2022-01-15
fix: #414 - use parent namespace to filter related resources

srfrnk:


grzesuav:
hi @srfrnk , this PR will introduce a regression, if the `parent` is cluster scope and you want to return related resource from any other namespace, it will override it. 


In your case, why your `customize` hook does not set the `Namespace` accordingly ?

srfrnk:
Good point. Please see new commit.

grzesuav:
Closing, https://github.com/metacontroller/metacontroller/issues/496 need to be implemented first, later on work on https://github.com/metacontroller/metacontroller/issues/345

Issue #302, opened 2021-07-06
Helm chart improvements


grzesuav:
Recently, thanks to @sathieu  we have helm chart, which can be improved in following way :

1. Add customization to arguments passed to metacontroller
2. When releasing, make chart bundle and attach to release resources as `assets` on https://github.com/metacontroller/metacontroller/releases page
3. Single source of truth for manifests, currently helm and manifests are separated, ideally manifests with different versions - `dev`/`debug`/`production` should be a result of helm chart generation with different input values happening on release time
4. CRD generation should point to helm chart once 3. is done

sathieu:
> 2. When releasing, make chart bundle and attach to release resources as assets on https://github.com/metacontroller/metacontroller/releases page

Having a chart index would be better. As an example, grafana [does this](https://github.com/grafana/helm-charts/blob/main/.github/workflows/release.yaml) using github pages.

grzesuav:
> > 1. When releasing, make chart bundle and attach to release resources as assets on https://github.com/metacontroller/metacontroller/releases page
> 
> Having a chart index would be better. As an example, grafana [does this](https://github.com/grafana/helm-charts/blob/main/.github/workflows/release.yaml) using github pages.

@sathieu can you post here a link to what chart index it ? I am noob in case of helm

grzesuav:
Actually I think they publish helm chart  release page - https://github.com/grafana/helm-charts/releases

sathieu:
@grzesuav The charts index is at https://grafana.github.io/helm-charts/index.yaml. It can be generated with `helm repo index`, see [here](https://helm.sh/docs/topics/chart_repository/) for generating it with shell script. I chose grafana charts as examples because they use github workflows.

grzesuav:
thanks, will look :+1:

sathieu:
See also https://github.com/marketplace/actions/helm-chart-releaser.

VictorBac:
I just found this thread, I made this earlier today because I needed it:
https://github.com/metacontroller/metacontroller/pull/364

This should correspond to your point number 1

orf:
This would be very useful, it's quite inconvenient to install this right now.

sathieu:
@orf Could you propose a PR here to implement this here?

orf:
Sure @sathieu, I've made #382.

grzesuav:
I closed #382 , it was using github pages to keep the charts. it is busy with serving documentation

grzesuav:
[...]

PR #420, opened 2022-01-15
chore(helm): #302 - add custom arguments to the helm chart

roelvdberg:
I have a usecase to be able to modify the metacontroller arguments when deploying it with the helm chart.

I have added the ability to add custom command line arguments to the helm chart. This way if the arguments change in the future we can keep the helm chart the way it is.


grzesuav:
hi @roelvdberg , can you :
* sign commit, DCO check has an instruction for it
* change message to `chore(helm): #302 - add custom arguments to the helm chart` (now it has typo in `chore` and missing `#` before issue number)

thank you

AmitKumarDas (approved):

PR #426, opened 2022-01-18
fix: change invalid log message when InPlace update strategy is used

vmax:
Currently, regardless of whether a recreate (`ChildUpdateRecreate`/`ChildUpdateRollingRecreate`) or inplace update (`ChildUpdateInPlace`/`ChildUpdateRollingInPlace`) strategy is selected for updating children, the same message (`Recreate update strategy selected`) is shown in the log. Instead, they should print different messages.


grzesuav (approved):

grzesuav:
thanks @vmax for this  PR!

PR #438, opened 2022-02-08
refactor: Use controller-runtime client to read crd's

grzesuav:


.github/config/kind-matrix-config.json
```diff
@@ -1,17 +1,5 @@
 {
   "configuration": [
-    {
```
AmitKumarDas, line 3: suggest: Can this be a separate commit?

go.mod
```diff
@@ -57,7 +59,6 @@ require (
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
-	k8s.io/apiextensions-apiserver v0.23.0 // indirect
```
AmitKumarDas, line 60: suggest: I guess go.mod needs to be formatted as a separate commit.

pkg/server/server.go
```diff
@@ -82,6 +84,10 @@ func New(configuration options.Configuration) (controllerruntime.Manager, error)
 	if err != nil {
 		return nil, err
 	}
+	err = apiextensionsv1.AddToScheme(mgr.GetScheme())
```
AmitKumarDas, line 87: suggest: Can we put some code comments on why the scheme needs to be added to v1alpha1 as well as apiextensionsv1.

AmitKumarDas (dismissed):
/LGTM
Have suggested a few changes

go.mod
grzesuav: no, I need to add apiextensions module to be able to handle CRD's, which are defined (schema and types) in that module

.github/config/kind-matrix-config.json
grzesuav: I need to drop this as it would not pass testing - I implemented `crdsv1` api version to get CRD from apiserver, in 1.14/1.15 it was in `v1beta1`  version.

pkg/server/server.go
grzesuav: added, I do not now why but CRD api is not in core module, but in `apiextensions`, so if I want to get CRD from api server I need to import that module and add to scheme

AmitKumarDas (approved):

PR #452, opened 2022-02-15
chore(registry): Publish to ghcr.io, make docker build parallel

grzesuav:


.github/workflows/release-pipeline.yaml
```diff
@@ -23,62 +30,31 @@ jobs:
...
-        context: .
-        file: ./Dockerfile
-        platforms: linux/amd64,linux/arm64,linux/arm
-        push: ${{ github.event_name != 'pull_request' }}
-        tags: ${{ steps.docker_meta_alpine.outputs.tags }}
-        labels: ${{ steps.docker_meta_alpine.outputs.labels }}
-        build-args: |
-          TAG=${{ env.GITHUB_REF_SLUG }}
-    - name: '[distroless] Docker metainformations'
+        registry: ghcr.io
```
AmitKumarDas, line 37: query: Is this a change in registry? What is the current registry?

AmitKumarDas: query: Should there be a doc update as well that mentions about the new registry?

AmitKumarDas (approved):
/LGTM
Left a few queries

.github/workflows/release-pipeline.yaml
grzesuav: @AmitKumarDas re: registry -  logging to both, and the name of the image determines to which ones image will be pushed, as I defined both names in `images` - the push will save to both.

I will add change to documentation separately, it should be also visible on `Packages` page

PR #485, opened 2022-03-28
feat(helm): implement pod disruption budget

roelvdberg:
Signed-off-by: Roel van den Berg <roel.vandenberg@infosupport.com>

I have implemented a Pod Disruption Budget for the controller pods so that we can guarantee that there is allways a pod online when running in a HA setup.

### DIFF
--- a/deploy/helm/metacontroller/templates/pdb.yaml
+++ b/deploy/helm/metacontroller/templates/pdb.yaml
@@ -1,0 +1,18 @@
+{{- if and .Values.podDisruptionBudget (gt (.Values.replicas | int) 1) }}
+apiVersion: {{ ternary "policy/v1" "policy/v1beta1" (semverCompare ">=1.21.0-0" .Capabilities.KubeVersion.Version) }}
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "metacontroller.fullname" . }}
+  labels:
+    {{- include "metacontroller.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "metacontroller.selectorLabels" . | nindent 6 }}
+    {{- if .Values.podDisruptionBudget.minAvailable }}
+    minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
+    {{- end }}
+    {{- if .Values.podDisruptionBudget.maxUnavailable }}
+    maxUnavailable: {{ .Values.podDisruptionBudget.maxUnavailable }}
+    {{- end }}
+{{- end -}}
--- a/deploy/helm/metacontroller/ci/pdb-value.yaml
+++ b/deploy/helm/metacontroller/ci/pdb-value.yaml
@@ -76,6 +76,7 @@ clusterRole:
       - "*"
 
 # podDisruptionBudget which can be enabled when running more then one replica's
-podDisruptionBudget: {}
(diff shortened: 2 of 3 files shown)


deploy/helm/metacontroller/values.yaml
```diff
@@ -64,3 +64,9 @@ cacheFlushInterval: 30m
 
 # The name of the PriorityClass that will be assigned to metacontroller
 priorityClassName: ""
+
```
mjsmith1028, line 67: I am thinking we can default this as:
```
podDisruptionBudget: {}
```

deploy/helm/metacontroller/templates/pdb.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if and .Values.podDisruptionBudget.enabled (gt (.Values.replicas | int) 1) }}
```
mjsmith1028, line 1: with the change to values.yaml, this could become
```
{{- if .Values.podDisruptionBudget }}
```
up to you if you also want to include the replica check, I don't have a strong opinion

mjsmith1028 (changes requested):
Thanks for the PR! I currently have PR open https://github.com/metacontroller/metacontroller/pull/483 to enable helm testing of leader election. I think this PR should merge after that PR so you add the values to `ci/leader-election-values.yaml` and it will lint/test the chart.

deploy/helm/metacontroller/templates/pdb.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: {{ ternary "policy/v1" "policy/v1beta1" (semverCompare ">=1.21.0-0" .Capabilities.KubeVersion.Version) }}
```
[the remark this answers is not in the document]
grzesuav: actually from where this version is taken ? From chart-file or some runtime option ?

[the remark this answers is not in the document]
roelvdberg: I will quote the helm documentation on this:
"Capabilities: This provides information about what capabilities the Kubernetes cluster supports."
https://helm.sh/docs/chart_template_guide/builtin_objects/

So this provides information about the runtime. We can't influence this from the chart.yaml

deploy/helm/metacontroller/values.yaml
roelvdberg: Thats also an option, I like having defaults in the chart.

deploy/helm/metacontroller/templates/pdb.yaml
roelvdberg: I prefer to have the replica check. 
The statefullset, which is used by metacontroller, won't update if it isn't allowed to bring at least one pod down.

A deployment can do a surge but a statefullset won't.

deploy/helm/metacontroller/templates/pdb.yaml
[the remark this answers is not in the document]
mjsmith1028: This is nice, I'm going to steal this on my PR 😄

grzesuav (dismissed):

grzesuav:
@roelvdberg can you rebase on latest master to get rid of conflicts ? Thank you

grzesuav (approved):

PR #499, opened 2022-04-08
refactor: #345 - Introduce UniformResourceMap

grzesuav:


pkg/controller/common/api/v1/common.go
```diff
@@ -130,7 +130,7 @@ func MakeRelativeObjectMap(parent v1.Object, list []*unstructured.Unstructured)
 }
 
 // GroupVersionKind is metacontroller wrapper around schema.GroupVersionKind
-// implementing encoding.TextMarshaler and encoding.TextUnmarshaler
+// implementing encoding.TextMarshaller and encoding.TextUnmarshaler
```
AmitKumarDas, line 133: nit: should this change be reverted

pkg/controller/common/api/v2/common.go
```diff
@@ -0,0 +1,147 @@
...
+		groupVersion.WithKind(parts[0]),
+	}
+	return nil
+}
+
+// UniformObjectMap holds objects grouped by GroupVersionKind and uniform location in the form namespace/name (if object is clustered scope, just name)
+// it holds entries in form [GroupVersionKind] -> [namespace/name] -> object
+// where namespace is empty in case of resources which are cluster scoped,
+// i.e. :
+// "v1/Pod -> 'prometheus/Prometheus'" (for namespaced child resource if the parent is cluster scope)
```
AmitKumarDas, line 50: nit: can this be improved?
e.g. Is below correct / better?
```sh
v1/Pod -> some-namespace/some-name -> Obj
```

pkg/controller/common/api/v2/common.go
```diff
@@ -0,0 +1,147 @@
...
+	return obj.GetName()
+}
+
+// FindGroupKindName search object by name in each schema.GroupKind (ignoring version part)
+func (m UniformObjectMap) FindGroupKindName(gk schema.GroupKind, name string) *unstructured.Unstructured {
+	// The map is keyed by group-version-kind, but we don't know the version.
+	// So, check inside any GVK that matches the group and kind, ignoring version.
+	for key, objects := range m {
+		if key.GroupKind() == gk {
+			for n, object := range objects {
```
AmitKumarDas, line 112: query: wondering if for loops can avoided since m is a map or maps

pkg/controller/common/api/v2/common.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+}
+
+// ReplaceObjectIfExists replaces the object with the same name & namespace as
+// the given object with the contents of the given object. If no object exists
+// in the existing map then no action is taken.
+func (m UniformObjectMap) ReplaceObjectIfExists(parent v1.Object, obj *unstructured.Unstructured) {
+	internalGvk := GroupVersionKind{obj.GroupVersionKind()}
+	objects := m[internalGvk]
+	if objects == nil {
```
AmitKumarDas, line 128: query: should we use below instead?
```go
objects, found := m[internalGvk]
if !found {
  return
}
```
```

**Review by mikesmithgh — CHANGES_REQUESTED** [review]

File: pkg/controller/common/api/v2/common.go (hunk 3 above)
Comment (grzesuav) [reply pkg/controller/common/api/v2/common.go:112]: hmm the inner one - I think yes. The outer - the key is GroupVersinKind, I need GroupKind, and I do not now Version

File: pkg/controller/common/api/v2/common.go (hunk 2 above)
Comment (grzesuav) [reply pkg/controller/common/api/v2/common.go]: yes, thanks

File: pkg/controller/common/api/v2/common.go [hunk 5]
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return nil
+}
+
+// UniformObjectMap holds objects grouped by GroupVersionKind and uniform location in the form namespace/name (if object is clustered scope, just name)
+// it holds entries in form [GroupVersionKind] -> [namespace/name] -> object
+// where namespace is empty in case of resources which are cluster scoped,
+// i.e. :
+// "v1/Pod -> 'some-namespace/some-name'" (for namespaced child resource)
+// "v1/Namespace -> 'some-namespace'" (for cluster-scope child resource)
```
AmitKumarDas, line 51: I guess this should be `some-name` instead of `some-namespace`

AmitKumarDas (approved):
/LGTM
Left one comment though

PR #505, opened 2022-04-09
fix(dynamic apply): Add `path` key as candidate to list merging

### LINKED ISSUE #443: Using downwardApi without apiVersion leads to never-ending pod delete/recreate loop
metacontroller version 2.2.0

Using the downwardAPI in a pod spec seems to lead to a never ending cycle of destroying/recreating the pod. Example:

```
downwardAPI:
 items:
   - path: pod_name
     fieldRef:
       fieldPath: metadata.name
```

A workaround is to add apiVersion: v1 under the fieldRef from our side.

grzesuav:
fixes #443

### DIFF
--- a/pkg/dynamic/apply/apply.go
+++ b/pkg/dynamic/apply/apply.go
@@ -212,6 +212,7 @@ var knownMergeKeys = []string{
 	"name",
 	"uid",
 	"ip",
+	"path",
 }
 
 // detectListMapKey tries to guess whether a field is a k8s-style "list map".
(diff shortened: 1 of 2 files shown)


pkg/dynamic/apply/apply.go
```diff
@@ -212,6 +212,7 @@ var knownMergeKeys = []string{
 	"name",
 	"uid",
 	"ip",
+	"path",
```
AmitKumarDas, line 215: suggest: It will be great to add unit tests for this change

AmitKumarDas (approved):

pkg/dynamic/apply/apply.go
grzesuav: hmm generic mechanism is tested in `apply_test` (example with `containerPort`) - I can add in separate PR test for checking what are know Merge Keys to catch up regressions

PR #508, opened 2022-04-11
feat(equality): Switch from reflect.DeepEqual to apiequality.Sematic.DeepEqual

grzesuav:


grzesuav:
it does not work as expected, seems that comparison isn't done based on types, probably because `unstructured` is used

PR #557, opened 2022-07-06
feature(webhooks): add etag support

dkropachev:
Implements https://github.com/metacontroller/metacontroller/issues/558

In order to reduce traffic between webhook server and metacontroller it worth to implement content caching.
With given implementation, i.e. POST methods, most suitable mechanism would he [eTagging](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag).

- eTag - a content hash that is supposed to be calculated of the content.
- It is getting exposed as a http header when webhook server responds to the client
- If client supports eTagging on next try it should send etag value of the content it cached as value of "If-None-Match" header.
- When webhook server sees this header and supports eTagging, it should send 304 "Unmodified" response with empty body if content (i.e. eTag) have not changed.
- Otherwise it sends full response as usual

It is all done in compliance with [rfc7232](https://datatracker.ietf.org/doc/html/rfc7232) section 2.3 and 3.2


dkropachev:
Only topic is not covered in this PR is cache invalidation, it is assumed that parent name, namespace, and kind are not dynamic values in the context of hook instance and therefore cache will not grow.
Please confirm if this assumption is correct, otherwise cache is subject to improvement.

grzesuav:
> Only topic is not covered in this PR is cache invalidation, it is assumed that parent name, namespace, and kind are not dynamic values in the context of hook instance and therefore cache will not grow. Please confirm if this assumption is correct, otherwise cache is subject to improvement.

hi @dkropachev , the children depends also on the related objects

pkg/etag_cache/etag_cache.go
```diff
@@ -0,0 +1,51 @@
+package etag_cache
```
grzesuav, line 1: I already use `zcache` for caching (example - https://github.com/metacontroller/metacontroller/blob/master/pkg/controller/common/customize/cache.go) with time-based eviction. Do you see any reason why not to use the same approach here ?

pkg/hooks/webhook_test.go
```diff
@@ -94,3 +151,136 @@ func TestWebhookTimeout_givenTimeoutIfPositive(t *testing.T) {
...
+		panic("unmarshal failed")
+	}
+	return &CallResult{
+		res: res,
+		err: err,
+	}
+}
+
+func TestHookETag(t *testing.T) {
+	body1 := "{\n  \"resyncAfterSeconds\": 30,\n  \"children\": [\n    {\n      \"apiVersion\": \"apps/v1\",\n      \"kind\": \"Deployment\",\n      \"metadata\": {\n        \"name\": \"res-name\"\n      },\n      \"spec\": {\n        \"key\": \"val\"\n      }\n    },\n    {\n      \"apiVersion\": \"apps/v1\",\n      \"kind\": \"Deployment\",\n      \"metadata\": {\n        \"name\": \"res-name2\"\n      },\n      \"spec\": {\n        \"key\": \"val2\"\n      }\n    }\n  ]\n}\n"
```
grzesuav, line 201: hmm maybe multi-line string will be more readable here ?

test/integration/framework/webhook.go
```diff
@@ -51,3 +51,49 @@ func (f *Fixture) ServeWebhook(handler func(request []byte) (response []byte, er
 	})
 	return srv
 }
+
+func (f *Fixture) ServeWebhook2(handler func(request []byte) (response []byte, headers map[string]string, err error), defaultHeaders map[string]string) *httptest.Server {
```
grzesuav, line 55: is it used anywhere ?

grzesuav (changes requested):

pkg/etag_cache/etag_cache.go
```diff
@@ -0,0 +1,51 @@
...
+type CacheEntry struct {
+	Etag        string
+	Response    []byte
+	LastUpdated time.Time
+}
+
+func NewCache() *Cache {
+	return &Cache{
+		cache: map[string]*CacheEntry{},
+		lock:  sync.RWMutex{},
```
AmitKumarDas, line 24: nit: Initialising is not required

pkg/hooks/webhook.go
```diff
@@ -99,15 +146,32 @@ func (w *WebhookExecutor) Call(request interface{}, response interface{}) error
...
 	}
 
-	// Check status code.
-	if resp.StatusCode != http.StatusOK {
+	if cacheEntry != nil {
+		// According to https://datatracker.ietf.org/doc/html/rfc7232
+		// When 'If-None-Match' is present, backend should respond 304 when object has not changed
+		if resp.StatusCode == 304 {
+			// TODO: Find a way to deep copy from cacheEntry.Response to response and switch cacheEntry.Response to store decoded response
+			if err = k8sjson.Unmarshal(cacheEntry.Response, response); err != nil {
```
AmitKumarDas, line 154: suggest: Let us avoid error shadowing. Let us use `if err := ...` instead i.e. `:=` instead of `=`

pkg/hooks/webhook.go
```diff
@@ -99,15 +146,32 @@ func (w *WebhookExecutor) Call(request interface{}, response interface{}) error
...
+			}
+			return nil
+		}
+	} else if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("remote error: %s", respBody)
 	}
 
 	// Decode response.
-	if err := k8sjson.Unmarshal(respBody, response); err != nil {
+	if err = k8sjson.Unmarshal(respBody, response); err != nil {
```
AmitKumarDas, line 164: Let us avoid err variable shadowing

pkg/hooks/webhook.go
```diff
@@ -99,15 +146,32 @@ func (w *WebhookExecutor) Call(request interface{}, response interface{}) error
...
 		return fmt.Errorf("can't unmarshal response: %w", err)
 	}
+
+	if cacheEntry != nil {
```
AmitKumarDas, line 168: suggest: This condition seems repeated & hence is prone to bugs & readability issues.
Can we try following:
```go
if cacheEntry == nil {
  k8sjson.Unmarshal(respBody, response);
}

// check for status code(s) here
```

AmitKumarDas: Please check if we can keep the condition for `resp.StatusCode != http.StatusOK` as-is & then add this new logic.
E.g.
```go
if resp.StatusCode != http.StatusOK {
  ...
}

if cacheEntry == nil {
  ...
}

// move the logic for status code 304 here
// logic to extract eTag
```

AmitKumarDas:
> Only topic is not covered in this PR is cache invalidation, it is assumed that parent name, namespace, and kind are not dynamic values in the context of hook instance and therefore cache will not grow. Please confirm if this assumption is correct, otherwise cache is subject to improvement.

IMO if the number of MetaController custom resources do not grow in an env then above assumption is ok. However, it might benefit if its invalidated when the custom resource gets deleted.

pkg/hooks/webhook.go
```diff
@@ -83,7 +119,18 @@ func (w *WebhookExecutor) Call(request interface{}, response interface{}) error
 		rawRequest := json.RawMessage(reqBody)
 		logging.Logger.Info("Webhook request", "type", w.hookType, "url", w.url, "body", rawRequest)
 	}
-	resp, err := w.client.Post(w.url, "application/json", bytes.NewReader(reqBody))
+
+	var cacheEntry = w.etagCache.Get(request.GetCacheKey())
```
AmitKumarDas, line 123: suggest: This can be move to later parts of the code. E.g. we do not need this entry that involves lock if response failed.

AmitKumarDas, line 123: important: etagCache can be nil & hence is prone to nil pointer error

pkg/etag_cache/etag_cache.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+func (c *Cache) Get(key string) *CacheEntry {
+	if c == nil {
+		return nil
+	}
+	c.lock.RLock()
```
AmitKumarDas, line 32: query: What is the reasoning behind a read lock. In other words, don't we need a write lock while setting the cache?

pkg/etag_cache/etag_cache.go
dkropachev: reimplemented in zcache

pkg/etag_cache/etag_cache.go
dkropachev: not relevant anymore

pkg/etag_cache/etag_cache.go
dkropachev: not relevant anymore

pkg/hooks/webhook.go
dkropachev: We need this entry before we run query, since if it is there we need to include If-none-match header with cached etag

dkropachev: it is handled under .Get method

pkg/hooks/webhook.go
dkropachev: done

pkg/hooks/webhook.go
dkropachev: done

pkg/hooks/webhook.go
dkropachev: > suggest: This condition seems repeated & hence is prone to bugs & readability issues. Can we try following:
> 
> ```go
> if cacheEntry == nil {
>   return k8sjson.Unmarshal(respBody, response);
> }
> 
> // move the logic for status code(s) here
> // logic to extract eTag
> ```

Nope, we do not want to cache broken json, so we store it after it is parsed

dkropachev: > Please check if we can keep the condition for `resp.StatusCode != http.StatusOK` as-is & then add this new logic. E.g.
> 
> ```go
> if resp.StatusCode != http.StatusOK {
>   ...
> }
> 
> if cacheEntry == nil {
>   ...
> }
> 
> // move the logic for status code 304 here
> // logic to extract eTag
> ```

Nope, we can't, reponse status 304 will fail `resp.StatusCode != http.StatusOK` condition.
And it is valid only when etag is enabled.

dkropachev:
> > Only topic is not covered in this PR is cache invalidation, it is assumed that parent name, namespace, and kind are not dynamic values in the context of hook instance and therefore cache will not grow. Please confirm if this assumption is correct, otherwise cache is subject to improvement.
> 
> IMO if the number of MetaController custom resources do not grow in an env then above assumption is ok. However, it might be good to invalidate the cache when the custom resource gets deleted. Some envs depend on metrics & observability & hence can benefit if cache logic can emit its count as a metric.

Added cache ivalidation by timeout, which is native to zcache

pkg/apis/metacontroller/v1alpha1/types.go
```diff
@@ -119,8 +119,9 @@ type Webhook struct {
 	URL     *string          `json:"url,omitempty"`
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
 
-	Path    *string           `json:"path,omitempty"`
-	Service *ServiceReference `json:"service,omitempty"`
+	EtagSupported *bool             `json:"etagSupported,omitempty"`
```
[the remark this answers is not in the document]
dkropachev: please revisit

pkg/hooks/webhook.go
dkropachev: in general, agree, it looks like spagetti-code, there is only one way to make it better is to split code in multiple methods and either call them on condition or have two different  Call methods one for when etag is enabled, another - disabled.
Wdt ? do you want me to go this route ?

pkg/hooks/webhook.go
```diff
@@ -63,27 +73,70 @@ func NewWebhookExecutor(
...
+	httpClient HttpClientInterface,
+) (*WebhookExecutor, error) {
+	if webhook == nil {
+		return nil, nil
+	}
+	url, err := webhookURL(webhook)
+	if err != nil {
+		return nil, err
+	}
+	if err != nil {
```
mjsmith1028, line 94: remove this nil check with log

pkg/hooks/webhook.go
```diff
@@ -63,27 +73,70 @@ func NewWebhookExecutor(
...
+		logging.Logger.Info(err.Error())
+	}
+
+	var eTagCache *etag_cache.Cache = nil
+	if webhook.Etag != nil && webhook.Etag.Enabled != nil && *webhook.Etag.Enabled {
+		eTagCache = etag_cache.NewCache(webhook.Etag.CacheTimeoutSeconds, webhook.Etag.CacheTimeoutSeconds)
+	}
+
```
mjsmith1028, line 102: should remove below nil check for err since it is checked above

pkg/hooks/webhook.go
```diff
@@ -63,27 +73,70 @@ func NewWebhookExecutor(
...
+		return err
+	}
+	if cacheEnabled && cacheEntryExists {
+		req.Header.Set("If-None-Match", cacheEntry.Etag)
+		eTagValue = cacheEntry.Etag
+	}
 	if logging.Logger.V(6).Enabled() {
 		rawRequest := json.RawMessage(reqBody)
-		logging.Logger.Info("Webhook request", "type", w.hookType, "url", w.url, "body", rawRequest)
+		logging.Logger.Info("Webhook request", "type", w.hookType, "url", w.url, "etag", eTagValue, "body", rawRequest)
```
mjsmith1028, line 136: should make level 6 instead of info `Logger.V(6)`

pkg/hooks/webhook_test.go
```diff
@@ -1,17 +1,75 @@
...
+	"io/ioutil"
 	"metacontroller/pkg/controller/common"
+	v11 "metacontroller/pkg/controller/composite/api/v1"
+	"metacontroller/pkg/logging"
+	"net/http"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
mjsmith1028, line 15: doesn't matter much because it is a test but could use  `metav1`

pkg/hooks/webhook_test.go
```diff
@@ -1,17 +1,75 @@
 package hooks
 
 import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
 	"metacontroller/pkg/controller/common"
+	v11 "metacontroller/pkg/controller/composite/api/v1"
```
mjsmith1028, line 8: doesn't matter much because it is a test but could use `compositev1`

mjsmith1028 (changes requested):
Added a few comments, thanks for the PR!

pkg/hooks/webhook.go
dkropachev: thanks, a leftover, removed

pkg/hooks/webhook.go
dkropachev: thanks, a leftover, removed

pkg/hooks/webhook.go
dkropachev: fixed

pkg/hooks/webhook_test.go
dkropachev: renamed

pkg/hooks/webhook_test.go
dkropachev: renamed

pkg/hooks/webhook.go
```diff
@@ -66,24 +76,61 @@ func NewWebhookExecutor(
...
+func (w *WebhookExecutor) Call(request WebhookRequest, response interface{}) error {
 	// Encode request.
 	reqBody, err := k8sjson.Marshal(request)
 	if err != nil {
 		return fmt.Errorf("can't marshal request: %w", err)
 	}
+
+	cacheKey := request.GetCacheKey()
+	cacheEnabled := w.etagCache != nil
+	cacheEntry, cacheEntryExists := w.etagCache.Get(cacheKey)
```
grzesuav, line 117: what if `w.etagCache` is nil here ?

pkg/hooks/webhook.go
```diff
@@ -66,24 +76,61 @@ func NewWebhookExecutor(
...
 	}
+	return NewWebhookExecutorWithCustomHttpClient(webhook, hookType, client)
+}
+
+func NewWebhookExecutorWithCustomHttpClient(
+	webhook *v1alpha1.Webhook,
+	hookType common.HookType,
+	httpClient HttpClientInterface,
+) (*WebhookExecutor, error) {
+	if webhook == nil {
```
grzesuav, line 87: this is tested in `NewWebhookExecutor` - method which is invoking this one. Actually this method should not be exported (is used internally only) and should not repeat logic which is in callermethod

pkg/hooks/webhook.go
```diff
@@ -66,24 +76,61 @@ func NewWebhookExecutor(
...
+		return nil, nil
+	}
+	url, err := webhookURL(webhook)
```
grzesuav, line 90: is is done in caller method, do not repeat that

grzesuav:
@dkropachev additionally, I see there is no documentation around that feature, would you mind add it as well to  https://metacontroller.github.io/metacontroller/api/hook.html

grzesuav:
> @dkropachev additionally, I see there is no documentation around that feature, would you mind add it as well to https://metacontroller.github.io/metacontroller/api/hook.html

is it here - https://github.com/metacontroller/metacontroller/blob/master/docs/src/api/hook.md

pkg/hooks/webhook.go
dkropachev: made it unexportable, removed webhook verification and url calculation from there

pkg/hooks/webhook.go
dkropachev: removed

dkropachev:
> > @dkropachev additionally, I see there is no documentation around that feature, would you mind add it as well to https://metacontroller.github.io/metacontroller/api/hook.html
> 
> is it here - https://github.com/metacontroller/metacontroller/blob/master/docs/src/api/hook.md

thanks, added

grzesuav (dismissed):
In general looks ok from my side, documentation should be improved to be more clear for the users. @mikesmithgh @AmitKumarDas please take a second look

grzesuav:
@dkropachev tests are failing, did you do `go mod tidy` in both main dir and `/test/integration` dir ? Seems missing deps

AmitKumarDas (dismissed):

dkropachev:
> @dkropachev tests are failing, did you do `go mod tidy` in both main dir and `/test/integration` dir ? Seems missing deps

thanks, run mod tidy, also found missing part of config-related code

grzesuav:
@dkropachev I do not see changes in CRD definition, did you run `make generate_crds` ?

Also, if all its done, please  rebase to single commit in order to have clean history

thank you !

dkropachev:
> @dkropachev I do not see changes in CRD definition, did you run `make generate_crds` ?
> 
> Also, if all its done, please rebase to single commit in order to have clean history
> 
> thank you !

Thanks, it is all done, cicd failures related to vulnerability checks, I have covered them in new [PR-568](https://github.com/metacontroller/metacontroller/pull/568)

All commits are squashed

grzesuav:
@dkropachev I do not see crd changes in https://github.com/metacontroller/metacontroller/blob/master/manifests/production/metacontroller-crds-v1.yaml, also left few nits in documentation

grzesuav:
@AmitKumarDas @mikesmithgh would you mind resolving comment if they are adjusted ? Thank you !

pkg/hooks/webhook.go
AmitKumarDas: ok. Lets keep your changes as-is.

AmitKumarDas (approved):
/LGTM

docs/src/api/hook.md
```diff
@@ -39,3 +40,25 @@ Within a `webhook`, the `service` field has the following subfields:
 | namespace | The `metadata.namespace` of the target Service. |
 | port | The port number to connect to on the target Service. Defaults to `80`. |
 | protocol | The protocol to use for the target Service. Defaults to `http`. |
+
+### Etag Reference
+
+More details in [rfc2616](https://datatracker.ietf.org/doc/html/rfc2616).
+
+Etag is a hash of endpoint content, backend that supports etag notion should issue "ETag" header to each 200 response.
+Client that supports "ETag" (in this particular instance) should send "If-None-Match" tag with value of ETag of cached content.
```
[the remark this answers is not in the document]
dkropachev: fixed

dkropachev:
> @dkropachev I do not see crd changes in https://github.com/metacontroller/metacontroller/blob/master/manifests/production/metacontroller-crds-v1.yaml, also left few nits in documentation

Thanks, doc comments addressed, also I figured that "make generate_crds" is broken due to the using outdated options, such as trivialVersion.

Crds updated.

mikesmithgh (approved):
LGTM 👍

grzesuav (approved):

Issue #496, opened 2022-04-06
[Feature Proposal] - Add versioning of HookRequest/Response api to Composite/Decorator controllers


grzesuav:
# Proposal

Add new field to `Decorator`/`Composite` CRD's in `Hook` field - to determine which request/response format should be used to communicate. 

Currently it looks like 
```yaml
  hooks:
    sync:
      webhook:
        url: http://cluster-parent-controller.metacontroller/sync
```
after change :
```yaml
  hooks:
    sync:
      version: v1
      webhook:
        url: http://cluster-parent-controller.metacontroller/sync
```

The field will be defaulted to `v1` (which is currently used request/response format).

The goal is to allow smooth introduction of changes which requires changing format of request/response (example - https://github.com/metacontroller/metacontroller/issues/345). Proposed FR was discussed in https://github.com/metacontroller/metacontroller/discussions/370.

## What will change
* new field in CRD - it is additive change, so backward compatible. Also we default to `v1` - current version

### `v2` proposal
* Change of `children`/`related` format from `RelativeObjectMap` to `UniformObjectMap` - in short - always have namespace scoped resources with namespace as a key - to make it unambiguous,
* `Decorator` controller sync/finalize request changes - [link](https://metacontroller.github.io/metacontroller/api/decoratorcontroller.html#sync-hook-request)
    *  rename `object` to `parent`
    * rename `attachements` to `children` 

This FR is about introducing way to handle different version. Implementation of version `v2` of rquest/response is outside of the scope of this FR.

## Challenges

1. Maintaining different versions of API - we should limit to i.e. two active versions, to not have burden.
2. Updating documentation
3. Testing

grzesuav:
https://github.com/metacontroller/metacontroller/pull/499

grzesuav:
When we have #499 and #564 next steps are :
* add conversion method to convert `UnformResourceMap` to `RelativeResourceMap`
* add handling of selected version in hooks
* add documentation

grzesuav:
Updated description with proposed changes to API in v2

mwgamble:
Is there any kind of timeline yet for deprecating and removing the v1 payload format?

grzesuav:
@mwgamble not for now, I plan to keep both for a while. v2 is still in progress though, can think about timeline for deprecating v1 when we will have established v2 and running for a while

mwgamble:
@grzesuav Is there anything I can do to help move this along? :pray:

[...]

PR #564, opened 2022-07-14
feat(hooks): Add versioning to hook API #496

grzesuav:


grzesuav:
need to add defaulting when not set

AmitKumarDas (approved):

PR #569, opened 2022-07-21
fix(security): Add ReadHeaderTimeout to pprof server to mitigate G112

grzesuav:


grzesuav:
@AmitKumarDas @mikesmithgh I will fix your comments in next PR, want to merge this to unblock release and it is pretty straighforward

PR #609, opened 2022-08-13
chore(release): Publish MacOs and Windows binaries

grzesuav:


.goreleaser.yaml
```diff
@@ -8,8 +8,8 @@ builds:
       - CGO_ENABLED=0
     goos:
       - linux
-#      - windows
-#      - darwin
+      - windows
```
mikesmithgh, line 11: any particular reason these were previously commented?

grzesuav: not really, just wanted to start with similar build metacontroller had in previous approach

mikesmithgh (approved):

PR #642, opened 2022-09-06
feat(webhooks): Select json deserialization mode of response: 'loose` (default) or 'strict'

### LINKED ISSUE #572: Customize hook accepts incorrect labelSelector format
The customize hook can return a list of related resources that are required for the sync hook.
Resources can be filtered using a labelSelector, however when the labelSelector is of the wrong format, the labelSelector basically becomes empty, but no warning or error is logged by the metacontroller.

Would be nice if an incorrect labelSelector would throw an error describing what is wrong. Alternatively, the docs could be updated to provide some examples.

grzesuav:
Adding possibility to set json deserialize mode to strict, which additionaly check for duplicated and unknown fields. Change is backward-compatibile, it defaults to 'loose'.

In addition, added few validations to CRD's, and used `testify` library.

Fixes #572

Signed-off-by: Grzegorz Głąb <grzesuav@gmail.com>

### DIFF
--- a/manifests/production/metacontroller-crds-v1.yaml
+++ b/manifests/production/metacontroller-crds-v1.yaml
@@ -46,6 +46,12 @@ spec:
                     updateStrategy:
                       properties:
                         method:
+                          enum:
+                          - OnDelete
+                          - Recreate
+                          - InPlace
+                          - RollingRecreate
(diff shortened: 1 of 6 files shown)


mikesmithgh (approved):
lgtm

PR #649, opened 2022-09-10
fix(etag): do not tollerate non-200 responses when cache is there

dkropachev:
When etag feature is turned on and cache is present it treats non 200 and non 403 responses as valid ones.
As result if server responds with 500 with empty body metacontroller will start to cleanup childrens.


grzesuav:
hi @dkropachev , can you open an issue for that ?

dkropachev:
This issue is already fixed on 3.4.0

Issue #650, opened 2022-09-13
[Bug] Mounting configMap to multiple paths detects diff.


brammel123:
If, in your webhook response, you specify a single configMap volume and mount this to several different locations in a single container, kubernetes will correctly mount them, but the metacontroller will detect a difference (even though there is none), and restart the resource indefinitely.

The "Diff between observed and desired" message reports a diff in the volumemounts, but specifies the same path 3 times.
Even though the webhook request lists them properly as 3 different paths.

If we create 3 seperate configmaps (same path, but different name) and mount those instead, the metacontroller behaves as expected.

I found a related [issue](https://github.com/GoogleCloudPlatform/metacontroller/issues/182) in the previous repo, that never got resolved.

**webhook response/request** (both the request and the response are identical)

        "volumeMounts": [
        {
            "name": "config",
            "mountPath": "/etc/a/extra.ini",
            "readOnly": true
        },
        {
            "name": "config",
            "mountPath": "/etc/b/extra.ini",
            "readOnly": true
        },
        {
            "name": "config",
            "mountPath": "/etc/c/extra.ini",
            "readOnly": true
        }]

**Diff between observed and desired** (notice thesame path occurs 3 times)

      "volumeMounts": [
          {
              "mountPath": "/etc/c/extra.ini",
              "name": "config",
              "readOnly": true
          },
          {
              "mountPath": "/etc/c/extra.ini",
              "name": "config",
              "readOnly": true
          },
          {
              "mountPath": "/etc/c/extra.ini",
              "name": "config",
              "readOnly": true
          }]
          
**Volume**

          {
            "name": "config",
            "configMap": {
                "name": "myname-config"
              }
          },

PR #661, opened 2022-09-17
fix: [#650] - Added 'mountPath' as possible merge-key to avoid that 'name' is treated as unique value for volumeMounts

EmielBruijntjes:
In issue #650 we ran into the problem that volumeMounts were not correctly merged. This happens when you have multiple volumeMounts that all refer to the same volume, but using different mountPaths. The reason for this seems to come from the merging-algorithm in the meta-controller, which is a bit fuzzy and incorrectly assumes that 'name' always has a unique value (which is not the case here, because 'name' refers to the volume and the same volume can be used for multiple volumeMounts).

In this pull-request we added an "mountPath" option as extra option to the list of possible mergeKeys. This fixes the issue for us, and I think it is reasonably safe to assume that the mountPath property is normally unique.

### DIFF
--- a/pkg/dynamic/apply/apply.go
+++ b/pkg/dynamic/apply/apply.go
@@ -209,6 +209,8 @@ func stringMergeKey(val interface{}) string {
 var knownMergeKeys = []string{
 	"containerPort",
 	"port",
+	"name",
+	"mountPath",
 	"uid",
 	"ip",
 	"path",
--- a/pkg/dynamic/apply/apply.go
+++ b/pkg/dynamic/apply/apply.go
@@ -209,7 +209,6 @@ func stringMergeKey(val interface{}) string {
 var knownMergeKeys = []string{
 	"containerPort",
 	"port",
-	"name",
 	"uid",
 	"ip",
 	"path",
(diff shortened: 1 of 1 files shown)


EmielBruijntjes:
Hmmmm, so difficult to get through the github merge process. It is only one line that was added to the code, do I really have to study to find out how this "codecov" robot and "signed-off-by" process works? I would be just as happy if this PR was rejected and that someone else who *does* know the rules and procedures of this repo would add this extra line, or tell me why our fix was wrong.

Is there anything that I can do to speed up the process? Maybe @grzesuav? It would be really nice if you can pick it up from here, and/or give some feedback.

grzesuav:
@EmielBruijntjes when you click `details` on DCO it tell you what to do

grzesuav:
also ,please squash the commits to single one, with `fix(#650) : Add 'mountPath' as merge key` as we use semantic commits

EmielBruijntjes:
Could you be so kind to help me out with this? I am not so experienced with git. I only know 'git commit' and 'git push' (sorry about that). If this was a 500 line change I would be more than happy to do all the studying, write unit tests and so on, but could you just take a look at our 1-line change and give feedback like "no, this stinks" or "yes, this is helpful" -- I prefer to fail fast here.

grzesuav:
if it helps then I am for it, ultimate solution for all the issues with merging will be switching to SSA, I can create separate PR with that if you are not advanced with git

grzesuav:
https://github.com/metacontroller/metacontroller/pull/663

EmielBruijntjes:
Great, awesome! I close this one then.

PR #662, opened 2022-09-18
feat: Default update strategy

zefir01:
This feature allows you to not explicitly specify child update strategies. You can specify a default update strategy.

If the default update strategy other than OnDelete is enabled, a ConfigMap will be created to store the GVKs of past applied children. Next, this ConfigMap will be read to query the list of children and create informers dynamically.


grzesuav:
@zefir01 can you create an issue with feature request when you describe which problem this commit address ?

PR #688, opened 2022-10-10
feat: Add api for trigger immediate reconciliation

timdevlet:
# Motivation

We've built our software atop of Metacontroller. Currently we have to wait for up to `resyncAfterSeconds` (30 in our case) for changes to be applied, but we'd like the system to be more responsive overall. One of the architectural improvements we've planned and implemented is out-of-order sync: Metacontroller would expose a simple HTTP endpoint that would trigger reconciliation on demand.

# Implementation details

* New API endpoint (disabled by default and is enabled by `--api=true` flag) `/trigger_sync` that listens on port 8090 and triggers immediate reconciliation
* GIN web framework was chosen due to performance and productivity reasons.


grzesuav:
hi @timdevlet , have some remarks/question to the PR :
1. Why not expose method on `metacontroller` for doing `sync` ? Exposing fields does seems good
2. Why GIN for such simple handler ? What does it gives ? How it will behave on two concurrent requests ?
3. It seems that `resync` is done by calling controllers `Start` method, is this right thing to do ? This method initializes whole structure. Additionally, is is done without calling `Stop` earlier

timdevlet:
Hello! I am still digging into project, so maybe I missed some architectural concepts. If you guide me and explain how to do this correctly I hope I can finalize this PR and you will be happy new functionality. 

1. I looked at this code and didn't find an elegant solution to reset queue timers. Or extract all possible keys to sync them in the loop. I can store them somewhere but it is a more dirty hack. 
```
func (pc *parentController) processNextWorkItem() bool {
       // Get all keys here <--------------------
	key, quit := pc.queue.Get()
	...
        // Loop keys and call pc.sync(key.(string)); <--------------------
	if err := pc.sync(key.(string)); err != nil {
		utilruntime.HandleError(fmt.Errorf("failed to sync %v '%v': %w", pc.parentResource.Kind, key, err))
		pc.queue.AddRateLimited(key)
		return true
	}
}
```

2. You are right. I use net/http but got CVE-2022-27664 error. GIN is an excellent but I used temporary solution to pass actions tests.
```
In net/http in Go before 1.18.6 and 1.19.x before
1.19.1, attackers can cause a denial of service
because an HTTP/2 connection can hang during
closing if shutdown were preempted by a fatal
error.
```
3. You mean I should stop before starting?
```
pController.Stop()
pController.Start()
```

grzesuav:
@timdevlet no worries, it just looked at first glance as a bit hacky solution. IMO we should not stop/start controllers. rather re-queue all objects - this is the way in which I would approach this

PR #697, opened 2022-11-01
feat(composite controller): Add label selector to choose parent objects

### LINKED ISSUE #695: What is the recommended upgrade procedure? Or: possible bug, "apiVersion" is not checked in "parentResource" of CompositeController
Is there documentation about the recommended procedure for (slowly) rolling out changes to webhooks?

For example, when we make changes to a webhook script, we rather not have such changes automatically applied when the periodic sync-process runs. So instead of making changes to a webhook, we prefer to add a *new* webhook, and use that for selected resources. This allows us to better keep an eye on possible errors. I tried to achieve this by setting up multiple webhook endpoints, one that is known to work, and a "beta" one we use for new code:
```
apiVersion: metacontroller.k8s.io/v1alpha1
kind: CompositeController
metadata:
  name: example-v1
spec:
  parentResource:
    apiVersion: example.com/v1
    resource: examples
  childResources: ...
  hooks:
    sync:
      webhook:
        url: http://webhook.metacontroller/examples/v1
---
apiVersion: metacontroller.k8s.io/v1alpha1
kind: CompositeController
metadata:
  name: example-v2alpha1
spec:
  parentResource:
    apiVersion: example.com/v2alpha1
    resource: examples
  childResources: ...
  hooks:
    sync:
      webhook:
        url: http://webhook.metacontroller/examples/v2alpha1
```
In the associated CRD we can then add a "v1" and a "v2alpha1" version. I thought that this would be a nice way to slowly rollout a new version. The "v2alpha1" endpoint will only be called for clients that explicitly use "example.com/v2alpha1".

However, when we create a "Kind: Example" resource, with apiVersion set to "example.com/v1", **both** of the above endpoints are called. This surprised me. Is this a bug in MetaController (that "parentResource.apiVersion" is not checked) or is my understanding of this process wrong, and is there an alternative process for rolling out changes to the webhook?

grzesuav:
fixes #695

Adds label selector to composite controller, which works the same way as in decorator controller. I decided not to add annotation selector, as using annotations for selections is against its design.

### DIFF
--- a/pkg/controller/composite/controller.go
+++ b/pkg/controller/composite/controller.go
@@ -25,6 +25,8 @@ import (
 	"sync"
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
 	"github.com/go-logr/logr"
 
 	"metacontroller/pkg/events"
@@ -67,6 +69,7 @@ type parentController struct {
 	dynClient      *dynamicclientset.Clientset
 	parentClient   *dynamicclientset.ResourceClient
 	parentInformer *dynamicinformer.ResourceInformer
+	parentSelector labels.Selector
 
 	revisionLister mclisters.ControllerRevisionLister
 
@@ -157,6 +160,14 @@ func newParentController(
 	if err != nil {
 		return nil, err
 	}
+	parentSelector := labels.Everything()
+	// for backward compatibility - if not set, handle all resources
+	if cc.Spec.ParentResource.LabelSelector != nil {
+		parentSelector, err = metav1.LabelSelectorAsSelector(cc.Spec.ParentResource.LabelSelector)
+		if err != nil {
+			return nil, err
+		}
+	}
 
 	pc = &parentController{
 		cc:             cc,
@@ -165,6 +176,7 @@ func newParentController(
 		childInformers: childInformers,
 		parentClient:   parentClient,
 		parentInformer: parentInformer,
+		parentSelector: parentSelector,
 		parentResource: parentResource,
 		revisionLister: revisionLister,
 		updateStrategy: updateStrategy,
@@ -304,6 +316,14 @@ func (pc *parentController) processNextWorkItem() bool {
 }
 
 func (pc *parentController) enqueueParentObject(obj interface{}) {
+	// If the parent doesn't match our selector, and it doesn't have our
+	// finalizer, we don't care about it.
+	if parent, ok := obj.(*unstructured.Unstructured); ok {
+		if !controllerutil.ContainsFinalizer(parent, pc.finalizer.Name) && pc.doNotMatchLabels(parent.GetLabels()) {
+			return
+		}
+	}
+
 	key, err := common.KeyFunc(obj)
 	if err != nil {
 		utilruntime.HandleError(fmt.Errorf("couldn't get key for object %+v: %w", obj, err))
@@ -360,6 +380,12 @@ func (pc *parentController) resolveControllerRef(childNamespace string, controll
 		// ControllerRef points to.
 		return nil
 	}
+	if !controllerutil.ContainsFinalizer(parent, pc.finalizer.Name) && pc.doNotMatchLabels(parent.GetLabels()) {
+		// If the parent doesn't match our selector and doesn't have our finalizer,
+		// we don't care about it.
+		return nil
+	}
+
 	return parent
 }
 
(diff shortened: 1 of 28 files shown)


pkg/controller/composite/controller.go
```diff
@@ -304,6 +312,14 @@ func (pc *parentController) processNextWorkItem() bool {
 }
 
 func (pc *parentController) enqueueParentObject(obj interface{}) {
+	// If the parent doesn't match our selector, and it doesn't have our
+	// finalizer, we don't care about it.
+	if parent, ok := obj.(*unstructured.Unstructured); ok {
+		if !controllerutil.ContainsFinalizer(parent, pc.finalizer.Name) && pc.parentSelector != nil && !pc.parentSelector.Matches(labels.Set(parent.GetLabels())) {
```
AmitKumarDas, line 318: query: `parent.GetLabels()` Does this imply all the labels present on the Parent needs to match to proceed further?
I think it will be nice to update the PR body with a sample YAML of now things will work. Probably you have already thought of doc updates.

pkg/controller/composite/controller.go
```diff
@@ -360,6 +376,12 @@ func (pc *parentController) resolveControllerRef(childNamespace string, controll
 		// ControllerRef points to.
 		return nil
 	}
+	if !controllerutil.ContainsFinalizer(parent, pc.finalizer.Name) && pc.parentSelector != nil && !pc.parentSelector.Matches(labels.Set(parent.GetLabels())) {
```
AmitKumarDas, line 379: Same comment

pkg/controller/composite/hooks.go
```diff
@@ -39,7 +41,14 @@ func (pc *parentController) callHook(
...
 	// called while the object is pending deletion.
-	if request.Parent.GetDeletionTimestamp() != nil && pc.finalizeHook.IsEnabled() {
+	//
+	// In addition to finalizing when the object is deleted, we also finalize
+	// when the object no longer matches our composite selector.
+	// This allows the composite to clean up after itself if the object has been
+	// updated to disable the functionality added by the decorator.
+	if pc.finalizeHook.IsEnabled() &&
+		(request.Parent.GetDeletionTimestamp() != nil ||
+			(pc.parentSelector != nil && !pc.parentSelector.Matches(labels.Set(parent.GetLabels())))) {
```
AmitKumarDas, line 51: nit: It might be better to have this condition split into multiples to improve readability & maintainability

pkg/controller/composite/controller.go
AmitKumarDas: nit: Does it makes sense to have following condition as a dedicated function in controllerutil? I find this getting repeated across multiple files.
```go
pc.parentSelector != nil && !pc.parentSelector.Matches(labels.Set(parent.GetLabels())
```

AmitKumarDas (commented):
Looks good. have you thought about adding test scenario that matches the issue #695

pkg/controller/composite/controller.go
grzesuav: @AmitKumarDas no,  just the selector need to match given set, so if selector is `app=aa`  and labels set is `{app=aa, other=bb}` it will match

pkg/controller/composite/hooks.go
grzesuav: done

pkg/controller/composite/controller.go
grzesuav: also, updated docs and example in `configmappropagation`

pkg/controller/composite/controller.go
```diff
@@ -304,6 +316,14 @@ func (pc *parentController) processNextWorkItem() bool {
 }
 
 func (pc *parentController) enqueueParentObject(obj interface{}) {
+	// If the parent doesn't match our selector, and it doesn't have our
```
AmitKumarDas, line 319: query: I am unsure of this. Hence asking. 
Is it fine to enqueue if:
1/ cond1=true _i.e. parent does not contain finalizer_ & cond2=false _i.e. parent labels match_ -> leading to enqueue
2/ cond1=false _i.e. parent contains finalizer_ & cond2=true _i.e. parent labels don't match_ -> leading to enqueue

AmitKumarDas (commented):
Thanks for the rework. I still have a doubt which I have posted in form of a review comment. Please have a look.

pkg/controller/composite/controller.go
grzesuav: I was basin on https://github.com/metacontroller/metacontroller/blob/master/pkg/controller/decorator/controller.go#L315-L320

re.1 it is true as it could be new object, so we might  add finalizer
re.2 it it is true, it means that i.e. object loose its labels and we might have some work to do during releasing the resource


at least up to my knowledge ;)

AmitKumarDas: Thanks!!!

AmitKumarDas (approved):
/LGTM

PR #718, opened 2022-11-17
refactor(#496): Extract WebhookRequest interface and builders

grzesuav:


pkg/controller/composite/api/common/interface.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"metacontroller/pkg/apis/metacontroller/v1alpha1"
+	"metacontroller/pkg/controller/common/api"
+	commonv1 "metacontroller/pkg/controller/common/api/v1"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type WebhookRequestBuilder interface {
```
AmitKumarDas, line 27: query: Do we really need an interface? In other words, variations may be handled as new methods to the Builder struct.

pkg/controller/decorator/api/common/interface.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"metacontroller/pkg/apis/metacontroller/v1alpha1"
+	"metacontroller/pkg/controller/common/api"
+	commonv1 "metacontroller/pkg/controller/common/api/v1"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type WebhookRequestBuilder interface {
+	WithController(controller *v1alpha1.DecoratorController) WebhookRequestBuilder
```
AmitKumarDas, line 28: Same comment.
Have we have thought having following methods against a struct & hence avoid interface completely
- WithDecoratorController(...)
- WithCompositeController(...)

AmitKumarDas (commented):
/LGTM
I have few comments though

pkg/controller/decorator/api/common/interface.go
grzesuav: It is not only about decorator/composite controller, there will be underlying implementation for `v1`/`v2` which will be different in the way how they construct the resource, and it is the main reason of introducing the interface.

Also, having `WithDecoratorController` and `WithCompositeController` bring complexity to the `build` method which I want to avoid

pkg/controller/composite/api/common/interface.go
grzesuav: it would make `build` method needed to handle all cases, it is not desired, I do not see disadvantage of using interface to avoid that

AmitKumarDas (approved):
/LGTM

PR #732, opened 2022-12-30
feat: optionally take managed control of child resources

lukebond:
When decorator and composite controllers create child objects in response to the user's sync hook, it appends an `ownerReferences` block to the metadata of all children/attachments. At present that block has the `controller` field set to `true`. This makes sense for most use-cases, the operator should be the managing controller for resources it creates.

I have a use-case in which I don't want that set to `true` because I'm creating objects that need to be owned and controlled by another controller: namely cluster-api resources. With the current functionality cluster-api will refuse to set its own ownership reference because there is already one there with `controller: true` (which is correct behaviour). I can see this applying to other cases, such as Crossplane or ACK or other cloud orchestration k8s controllers.

If the user doesn't set this new field it will behave as it does now. If they set it to true it will behave as it does now. So it is opt-in.

I suspect that setting this field to false might have unexpected side-effects that confuse the kind of operator best-practices that Metacontroller strives to enforce, but since I'm new to the codebase it's a bit difficult for me to know the impact.

Please advise on any integration or end-to-end tests that I could add in order to test it thoroughly. I really didn't know where to start. I just added some very basic tests.

I have tested this in my setup as follows:
- rebuilt the metacontroller binary
- built a dev docker image
- pushed it to my docker hub
- hot edited my metacontroller stateful set to use my docker hub image
- deleted and re-applied the modified CRD manifests (the ones changed in this PR)
- modified my running compositecontroller, setting `managingController` to `false`
- created a new custom resource that is the parent for my operator

cluster-api was then happy to take ownership of those resources and drive the cluster to provisioned state. It works!

I have optimistically updated the docs, too.


docs/src/api/compositecontroller.md
```diff
@@ -251,6 +252,21 @@ new ones from scratch.
...
+owned by the parent object, which is best practice when building operators.
+There are some occasions, however, where you may not want Metacontroller to set
+the `controller` field of this `ownerReferences` object to `true`. This is
+usually when you are creating children that will become owned by some other
+controller, for e.g. a cluster-api cluster custom resource; such a resource
+needs to be owned and controller by its own controller.
+
+If you set this to `false`, the children will still get an entry in
+`ownerReferences` pointing to the parent, but the `controller` field will be
+set to `false`. It is not possible to set this value for some children and not
```
lukebond, line 267: after writing this i think it would be better to set this value per-resource, in the attachments array for decorators and the children array for composites.

lukebond:
i have refactored this PR to allow the `managingController` flag to be set for each individual child resource kind and updated the docs and tests accordingly.

i have added some integration tests to ensure that this mode doesn't mess with the cascading delete semantics:
- duplicated the existing "cascading delete" integration test, but with `managingController` set to `true` for the child resource (Job); it works
- wrote a new test to delete the parent CR (whereas the existing cascading delete test would update the replica count in the parent) and check that the child gets deleted. i added two versions of this test, with both possible values for `managingController`. i was surprised to find that both tests fail, suggesting a misunderstanding on my part about the relationship between the parents and its children

questions: would others expect that deleting the parent CR would cause any children it creates to be deleted? is it simply my naive sync function that always returns a resource? i understood that, without a finalise hook, all children would just be deleted right away. with some logging i was able to identify that my sync hook wasn't called again.

grzesuav:
> questions: would others expect that deleting the parent CR would cause any children it creates to be deleted? is it simply my naive sync function that always returns a resource? i understood that, without a finalise hook, all children would just be deleted right away. with some logging i was able to identify that my sync hook wasn't called again.

so if you delete parent resource, children should be deleted by garbage collector short after (at least when there is no other `ownerRefs` afaik and `blockDeletion` is `true`), no need to call finalize hook

pkg/controller/common/manage_children.go
```diff
@@ -113,13 +113,13 @@ func revertField(newObj, orig *unstructured.Unstructured, fieldPath ...string) e
...
 
-func MakeControllerRef(parent *unstructured.Unstructured) *metav1.OwnerReference {
+func MakeControllerRef(parent *unstructured.Unstructured, controller bool) *metav1.OwnerReference {
 	return &metav1.OwnerReference{
 		APIVersion:         parent.GetAPIVersion(),
 		Kind:               parent.GetKind(),
 		Name:               parent.GetName(),
 		UID:                parent.GetUID(),
-		Controller:         pointer.Bool(true),
+		Controller:         pointer.Bool(controller),
```
grzesuav, line 122: I am still unsure if this is correct, looking at https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L59-L120 it seems that they set `blockOwnerDeletion` along with `controller`.

grzesuav: reading https://github.com/kubernetes/design-proposals-archive/blob/main/api-machinery/controller-ref.md this field was introduced to avoid fighting over resources, so metacontroller would be creating object which it do not suppose to manage further

lukebond:
read this in the proposal link that was posted in a comment above:

---

- This proposal does not change how cascading deletion works.
  
  Although ControllerRef will extend OwnerReference and rely on its machinery, the Garbage Collector will continue to implement cascading deletion as before. That is, the GC will look at all OwnerReferences without caring whether a given OwnerReference happens to be a ControllerRef or not.

---

to me this suggests that setting the controller bool value shouldn't interfere with GC and cascading deletion (i still have to fix that test, and write one for the decorator, but it should work once i get it right).

lukebond:
i think this PR is ready for review now @grzesuav, although one e2e test fails, "clusteredparent/test.sh". i really cannot understand why my code would cause this to fail. all this PR does is plumb through one boolean and it isn't set by default. but looking at the PR history this test is not a flake. could you help me fix this test?

otherwise, i have removed the tests i added to test that deletion of the parent triggers deletion of the children. that does work in metacontroller, of course, but there weren't any tests for it specifically. it's internal k8s machinery rather than something we need to test here, i think. my tests didn't work anyway, and i don't know why (the children didn't get deleted even without my managing controller change). i tried that test against master branch and it fails there too, so it's not that i've broken the deletion behaviour for when managingController is true.

i did test that this deletion works in my cluster running a build of metacontroller from this branch, here's what i did:
- created a jsonnet controller that outputs a job from a CR parent
- configured that controller with one job child with `managingController` set to `false`
- created a CR, watched the job get created, then deleted the CR. the job gets deleted as expected

there is a trivy scan failure but i didn't introduce anything new; is there anything i should do about this?

for me the only outstanding question is this one, from a comment above, now resolved:

>> (@grzesuav) so metacontroller would be creating object which it do not suppose to manage further
>
>(@lukebond) yes that's essentially what i'm trying to do. create objects that will be managed by someone else, but still benefit from cascading deletion. i want the creation of a CR to trigger the creation of other objects, but the controller i write to create them won't be managing them. reading this design doc it does seem to violate the "three laws", which explicitly states that you should take a ControllerRef.
>
>do you think this is something best done with a custom operator and not built into metacontroller? it seems like a pretty basic thing to do and seems a shame to not be able to leverage metacontroller for the operator control loops etc. that it does for you.

for me it seems harmless as it's opt-in and not the default, and allows those like me who explore this to verify if it works in the wild :shrug:

lukebond:
i am closing this PR as the problem in my use-case is not with metacontroller, it's with cluster-api taking a controller ref on resources it doesn't create. [i think that should change](https://github.com/kubernetes-sigs/cluster-api/issues/4014) but in the meantime for our use-case we will roll our own controller that doesn't take controller ref, against operator best practice.

PR #733, opened 2022-12-30
docs: some clarifications to build and test docs

lukebond:
Signed-off-by: Luke Bond <luke.n.bond@gmail.com>

In the process of making a PR for this project I found some things unclear or missing in the docs. This PR addresses the issues I found.

- It wasn't clear how to compile the binary, therefore I assumed the Docker build would compile it
- Some dependencies were needed for the build that I didn't have installed and took me a while to find
- Path location of the integration test binaries installation script at first appeared wrong but then I found it in a subdir
- The e2e tests don't require an empty cluster, as it states, but one that is already set up with metacontroller

### DIFF
--- a/docs/src/contrib/build.md
+++ b/docs/src/contrib/build.md
@@ -15,6 +15,20 @@ git clone git@github.com:metacontroller/metacontroller.git metacontroller
 cd metacontroller
 ```
 
+Then you can build a `metacontroller` binary like so:
+
+```sh
+make build
+```
+
+Note that you will need the following k8s code-generation tools:
+* deepcopy-gen
+* client-gen
+* lister-gen
+* informer-gen
+
+...all of which can be found [here](https://github.com/kubernetes/code-generator).
+
 ## Local build and development
 
 Check [debug section](./debug.md)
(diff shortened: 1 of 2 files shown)


docs/src/contrib/debug.md
```diff
@@ -12,6 +12,8 @@ There are different flavours of manifests shipped to help with local development
 
 ### Development build
 
+You must first build the binary with `make build`, the following steps will then add it to the Docker image.
+
 The main difference it that image defined in manifest is `metacontrollerio/metacontroller:dev`, therefore:
 * apply dev manifests - `kubectl apply -k manifests/dev`
 * build docker image with command - `docker build -t metacontrollerio/metacontroller:dev -f Dockerfile .`
```
grzesuav, line 19: actually currently it is done by `make image` step - https://github.com/metacontroller/metacontroller/blob/master/Makefile#L44

docs/src/contrib/build.md
```diff
@@ -15,6 +15,20 @@ git clone git@github.com:metacontroller/metacontroller.git metacontroller
 cd metacontroller
 ```
 
+Then you can build a `metacontroller` binary like so:
+
+```sh
+make build
+```
+
+Note that you will need the following k8s code-generation tools:
```
grzesuav, line 24: it will be automatically downloaded by Makefile - https://github.com/metacontroller/metacontroller/blob/master/Makefile#L73

grzesuav: it wasn't in your case ?

lukebond: ah it was put in my $GOPATH which is not in my path (i don't normally install things this way, preferring to only install things using my OS package manager). i'll remove this as i will be in the minority here for sure

grzesuav (approved):

grzesuav:
@lukebond thanks for this PR !

PR #745, opened 2023-01-25
refactor(#496) - Add v2 builders for composite controllers

grzesuav:


mikesmithgh (approved):
Looks good to me. It has been a little while since I have been in the metacontroller code so I just want to double check my understanding.

In our hooks definition, we can optionally specify a version. If we do not specify a version then it will default to `v1` which means we will use the new `v2/common` `Convert` function to make the webhook call backward compatible for existing controllers. So, if the user does nothing, then everything should behave like normal.

If the user updates the version to `v2` in the hooks definition, then they will have to update their controller for `UniformObjectMap` which will now let us just use a name for a cluster scoped resource.

Does that sound right to you?

AmitKumarDas (approved):
/lgtm

PR #750, opened 2023-02-08
fix: Support testing on arm64

mwgamble:
gotestsum provides pre-compiled binaries for arm64, but the script was previously hard-coded to only download amd64 binaries. Now it will detect the correct architecture to use before downloading.

I also took this opportunity to surround an instance of `${PWD}` in quotes, so that it's safer and more consistent with the rest of the script.

Signed-off-by: Matthew Gamble <matthew.gamble@rea-group.com>

### DIFF
--- a/test/integration/hack/setup.sh
+++ b/test/integration/hack/setup.sh
@@ -14,7 +14,16 @@ else
     echo ""
     echo "+++ downloading gotestsum"
     echo ""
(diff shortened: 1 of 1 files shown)


grzesuav (approved):

grzesuav:
@mwgamble thanks for this !

Issue #533, opened 2022-05-24
Feature request - Support for rate limiting (HTTP 429)


adrianmace:
## Summary

Our sync hook sends 429 responses back to Metacontroller when it is overloaded - in an effort to mitigate 3rd party API limits being exceeded - but it doesn't seem like Metacontroller is honouring it.

## Request

Could we please see that HTTP 429 responses are correctly handled? This should also include support for the [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header (which we set in attempt to control back-off).

## Example Error Log Entry

(Do note that we have forked metacontroller internally and modified the source to meet certain security desires, so the line numbers represented here may not be exact.)

```
{"level":"error","ts":1653018362.3708982,"msg":"failed to sync CUSTOM_RESOURCE 'NAMESPACE/METADATA_NAME': sync hook failed for CUSTOM_RESOURCE NAMESPACE/METADATA_NAME: sync hook failed: remote error: <!DOCTYPE HTML PUBLIC \"-//W3C//DTD HTML 3.2 Final//EN\">\n<title>429 Too Many Requests</title>\n<h1>Too Many Requests</h1>\n<p>Rate Limit exceeded: 1 per 1 second</p>\n","stacktrace":"metacontroller/pkg/controller/composite.(*parentController).worker\n\t/go/src/metacontroller/pkg/controller/composite/controller.go:286\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.23.5/pkg/util/wait/wait.go:155\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.23.5/pkg/util/wait/wait.go:156\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.23.5/pkg/util/wait/wait.go:133\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/go/pkg/mod/k8s.io/apimachinery@v0.23.5/pkg/util/wait/wait.go:90\nmetacontroller/pkg/controller/composite.(*parentController).Start.func1.1\n\t/go/src/metacontroller/pkg/controller/composite/controller.go:262"}
```

grzesuav:
[...]

PR #760, opened 2023-03-06
Support rate limiting to handle HTTP 429

weijiany:
# Support rate limit

Add a new feature to requeue item when the backend service returns the 429 HTTP code. The requeue after some seconds is based on `Reter-after` field in the response header.
Close: #533

### DIFF
--- a/pkg/hooks/webhook.go
+++ b/pkg/hooks/webhook.go
@@ -21,12 +21,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"metacontroller/pkg/cache"
 	"metacontroller/pkg/controller/common"
 	"metacontroller/pkg/controller/common/api"
 	"metacontroller/pkg/logging"
 	"metacontroller/pkg/metrics"
 	"net/http"
+	"strconv"
 	"time"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -98,6 +100,7 @@ func NewWebhookExecutor(
 		hookType,
 		webhook.ResponseUnmarshallMode,
 		abstract,
+		time.Now,
 	), nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/hooks/webhook.go
```diff
@@ -159,6 +165,18 @@ func (w *webhookExecutor) Call(webhookRequest api.WebhookRequest, webhookRespons
 	}
 	defer response.Body.Close()
 
+	if response.StatusCode == 429 {
+		var afterSecond int
+		retryAfter := response.Header.Get("Retry-After")
+		nextTime, err := time.Parse(time.RFC1123, retryAfter)
+		if err != nil {
+			afterSecond, _ = strconv.Atoi(retryAfter)
```
grzesuav, line 173: @weijiany how it will behave if there is unparsable value ?

grzesuav:
@weijiany thanks for the PR, left one question

pkg/hooks/webhook.go
weijiany: The value of `afterSecond` will be 0. I could set a default value for this, but I don't want to handle the error that is parsed. Because the HTTP protocol makes us if the response code is 429, the `Retry-After` only can support two kinds.

grzesuav: got it, so basically it will behave as when no option set, got it.

grzesuav (changes requested):
@weijiany can you fix linter error (complains on file not `goimported`) and rebase onto master (to fix security warning)?


Thanks

grzesuav:
Actually there is more 
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.52.0-linux-amd64/golangci-lint run --out-format=github-actions --timeout 10m --enable errorlint,gofmt,goimports,gosec,whitespace,bodyclose,dogsled,durationcheck,errorlint,exhaustive,exportloopref,goconst,gocritic,gosec,importas,misspell,nilerr --exclude=G108] in [] ...
  Error: File is not `goimports`-ed (goimports)
  Error: File is not `goimports`-ed (goimports)
  Error: type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors (errorlint)
  Error: composites: metacontroller/pkg/hooks.TooManyRequestError struct literal uses unkeyed fields (govet)
  
  Error: issues found
```

weijiany:
@grzesuav Could you run the github action to check again? I think I fix it. 😄

grzesuav:
@weijiany yes, thank you, last time it was leaving an appropriate comments on code which was helpful, not sure it stopped, will look into this

weijiany:
> have you run `goimports -w . ` ?

There are a lot of commits to fix linting issue, it becomes very verbose, apologize for that @grzesuav .

grzesuav:
no worries, I noticed that golangci-lint github action not not send `annotations`, not sure why - https://golangci-lint.run/usage/install#github-actions

grzesuav (approved):

PR #764, opened 2023-03-14
chore(docs): `-a` is not a vaild flag for kubectl get pods

fernandoataoldotcom:
kubectl get pods will take -A, which indicates --all-namespaces and is not needed in this case because the command declares the namespace


grzesuav (approved):

grzesuav:
@fernandoataoldotcom thanks for the pr

Issue #766, opened 2023-03-16
Status field not being updated for the attachment resource


Shashankft9:
I have noticed when I use a `DecoratorController`, and try to send the response with status for an attachment resource, although it gets reflected in the attachment resource for the very first time, but for the subsequent updates in the status field of this attachment resource, metacontroller just ignores doing any changes - its even missing in the mergepatch calculated.

sample logs from metacontroller:
in request i see this: (expected)
```
"status":{"conditions":[{"lastTransitionTime":"2023-03-06T10:23:53Z","status":"True","type":"CUD"}]}}}},"related":{},"finalizing":false}}
```
in response i see this: (expected)
```
"status":{"conditions":[{"type":"CUD","status":"True","lastTransitionTime":"2023-03-09T07:31:52Z"}]}}]}}
```
Notice the time change in the request and response.
Now, in this log, metacontroller gives the expected change in spec in attachment cr, but I cant see the status bit here
```
{"level":"Level(-5)","ts":1678347112.267575,"msg":"Diff between observed and desired","mergePatch":{"spec":{"data":"{\"maxclients\":\"9800\",\"repl-ping-replica-period\":\"10\"},,\"runAsNonRoot\":true,\"runAsUser\":7777777}}"}}}
```

grzesuav:
hi @Shashankft9 , I saw https://github.com/metacontroller/metacontroller/pull/776, however this is not the way how I would like to handle it, as it is arbitrary changing the behawior. In some cases  people do not controll the races in resources.


I would see that it will be controllable behaviour per child/attachement, so change in the config API, I still need to wrap an idea regarding this, as have some other config related changes in API types (i.e. emitting events)

PR #776, opened 2023-04-03
refactor: #766 - allowing attachment status change

Shashankft9:
This PR simply allows `status` changes in an `attachment` resource when using `DecoratorController`

More context: https://kubernetes.slack.com/archives/CA0SUPUDP/p1678098438458439?thread_ts=1677578734.006459&cid=CA0SUPUDP

cc @grzesuav


grzesuav (changes requested):
This is not how I would see the resolution, elaborated in #766

PR #839, opened 2023-08-14
feat: #218 - Add target-label-selector

### LINKED ISSUE #218: Feature proposal : Switch from shared to standalone model

newtondev:
Adds functionality to be able to run multiple instances of Metacontroller that are responsible for managing subsets of controllers (CompositeController/DecoratorController).

Documentation has also been added in the PR for this feature.

image

Fixes #218


newtondev:
> @newtondev can you locally run one of the `e2e` tests to check what is happening ?

I tracked down and fixed the problem, however I will need to write some additional tests to increase the coverage.

pkg/cmd/metacontroller/main.go
```diff
@@ -56,6 +56,7 @@ var (
 	leaderElectionNamespace    = flag.String("leader-election-namespace", "", "Determines the namespace in which the leader election resource will be created")
 	leaderElectionID           = flag.String("leader-election-id", "metacontroller", "Determines the name of the resource that leader election will use for holding the leader lock")
 	healthProbeBindAddress     = flag.String("health-probe-bind-address", ":8081", "The address the health probes endpoint binds to")
+	targetLabelSelector        = flag.String("target-label-selector", "", "Targeted labels for label selector used to limit metacontroller to watch for specific resources")
```
grzesuav, line 59: @newtondev can we emphasis it is for Composite/Decorator controllers ?

pkg/controller/composite/metacontroller.go
```diff
@@ -120,6 +125,19 @@ func (mc *Metacontroller) Reconcile(ctx context.Context, request reconcile.Reque
 			"[%s] Sync error - %s", cc.Name, err)
 		return reconcile.Result{}, err
 	}
+
+	// Check to see if this metacontroller instance manages the specified
+	// composite controller resource based on targeted labels.
+	if mc.targetLabelSelector != nil && !mc.targetLabelSelector.Empty() {
```
grzesuav, line 131: @newtondev as an alternative, have you considered doing that in predicate ? It would skip the reconcile logic entirely  - i.e. in `server.go` line 119 and 130

pkg/server/server.go
```diff
@@ -91,12 +93,23 @@ func New(configuration options.Configuration) (controllerruntime.Manager, error)
 		return nil, err
 	}
 
+	// If configuration has a TargetLabelSelector, then read the values
+	// and create the labels.Selector object.
+	var targetLabelSelector labels.Selector
+	logging.Logger.Info("Configuration targetLabelSelector text is: " + configuration.TargetLabelSelector)
```
grzesuav, line 99: we use structured logging, correct form would be 
```
logging.Logger.Info("Initializing target label selector", "label_selector", configuration.TargetLabelSelector)
```
or similar

pkg/server/server.go
```diff
@@ -91,12 +93,23 @@ func New(configuration options.Configuration) (controllerruntime.Manager, error)
...
+	if configuration.TargetLabelSelector != "" {
```
grzesuav, line 100: Actually I would see behavior here as
1. Initialize `labelSelector` as `labels.Everything()`
2. If there is `targetLabelSelector` set, override with the setting
I guess it would simplify the flow

```
targetLabelSelector := labels.Everything()
if (configuration.targetLabelSelector !="") {
  // read from config
}
```

pkg/server/server_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+const errShouldBeNil = "err should be nil, got: %v"
+
+// see: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
+func TestTargetLabelSelectorParsing(t *testing.T) {
+	var err error
+	// Single
+	_, err = labels.Parse("foo=bar")
```
grzesuav, line 32: we should not test `labels.Parse` method, I believe it is tested by upstream

grzesuav (changes requested):
I see some room for improvement here.

please do not mind codecov test, I am ok with slightly decreasin the coverage.

what would be nice is some `e2e` test for that

pkg/controller/composite/metacontroller.go
newtondev: Thanks @grzesuav . I will take a look at this next.

pkg/server/server.go
newtondev: Thanks @grzesuav . I have updated accordingly based on your suggestion.

pkg/cmd/metacontroller/main.go
newtondev: Thanks @grzesuav . I have updated accordingly based on your suggestion.

pkg/server/server.go
newtondev: Thanks @grzesuav . I have updated this to use structured logging as you suggested.

pkg/server/server_test.go
newtondev: Thanks @grzesuav . I have removed the tests for this. I initially added them just to test my logic, but they don't test the code path. I will look to adding some e2e tests.

pkg/controller/composite/metacontroller.go
newtondev: Added predicate filter function and passed to the watchers for both Composite and Decorator controllers. 

I will write some e2e tests for this next.

grzesuav:
Thanks @newtondev looks good now, would be nice to have e2e test (can be an ajdusted copy paste of one of the existing ones just to make sure it won't regress)

I am on fixing the release script, once its done I can merge this

newtondev:
Thanks @grzesuav . I added some e2e tests with some technical difficulties along the way. Looking forward to the fixes on the release script you are adding.

.release.sh
```diff
@@ -2,4 +2,4 @@
 set -euxo pipefail
 
 printf "%s" "$1" > /tmp/release-notes.md
-goreleaser release --release-notes /tmp/release-notes.md --rm-dist
```
grzesuav, line 5: why changing `-rm-dist` to `-clean` was needed ?

grzesuav:
@newtondev leave one question re changing gorelaser option

side comment - I see you write functional tests - which is also good, however they can be tricky to write (I have an open issue to migrate to controller-runtime envtest instead) - e2e test are in example directory (each of have subfolder with test name and `.test.sh` script) - they tend to me easier to write

grzesuav:
also, would you mind squashing commits to single one once all is done ? Like explained in [freecodecam](https://www.freecodecamp.org/news/git-squash-commits/) ? Thanks !

.release.sh
newtondev: It mentioned in the output that `--rm-dist` is deprecated and should be replaced with `--clean`. I can revert this change.

grzesuav (approved):

PR #843, opened 2023-08-18
fix(hooks): Allow cluster resources in related result

### LINKED ISSUE #842: Regression in 4.10.3 and up related to the customize hook
Hi,

Metacontroller version 4.10.3 has a regression compared to 4.10.2 when it comes to how the customize hook works.

One of our controllers is a decorator controller that acts on Deployment. In order to generate the child resource, it requires data (labels) from the namepace the Deployment is in. Our code contains this simple code:

```
    def customize(self, namespace: str) -> dict:
        return [
            {
                "apiVersion": "v1",
                "resource": "namespaces",
                "names": [namespace],
            }
        ]
```

When we run this against 4.10.2, the namespace data is returned correctly. When we run the same code against 4.10.3 or up, no namespace data is returned. No changes were done to our python code.

I've tried to reproduce it with the SecretPropagation example, but that seems to not be affected. Perhaps it is related to the parent resource being a cluster scoped resource instead of a namespace scoped one?

4.10.3 was released with some changes to support the v2 API for the customize hook, so it's probably related to that.

newtondev:
Fixes #842

### DIFF
--- a/pkg/controller/common/api/v2/common.go
+++ b/pkg/controller/common/api/v2/common.go
@@ -112,18 +112,19 @@ func (m UniformObjectMap) List() []*unstructured.Unstructured {
 }
 
 // Convert returns commonv1.RelativeObjectMap against given parent, removing non matching objects
-func (m UniformObjectMap) Convert(parent *unstructured.Unstructured) commonv1.RelativeObjectMap {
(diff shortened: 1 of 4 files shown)


newtondev:
This could potentially cause data to leak unintentionally. The main use case for this is to get a list of the namespaces in a customize hook to return the labels and annotations thereof when using a Namespaced controller. I could update the code to allow for that as the unintentional behavior introduced with this change, could potentially get sensitive information like secrets.

PR #851, opened 2023-08-23
fix(hooks): Make ETag header RFC compliant

### LINKED ISSUE #850: ETag Support Is Not RFC Compliant
According to [RFC7232](https://datatracker.ietf.org/doc/html/rfc7232#section-3.2) (which obsoletes [RFC2616](https://datatracker.ietf.org/doc/html/rfc2616#section-14.26)), `ETag` support, specifically when responding to requests that contain a `If-None-Match` header, is described as:

> An origin server MUST NOT perform the requested method if the
   condition evaluates to false; instead, the origin server MUST respond
   with either a) the 304 (Not Modified) status code if the request
   method is GET or HEAD or b) the 412 (Precondition Failed) status code
   for all other request methods.

[Webhook support for ETags](https://metacontroller.github.io/metacontroller/api/hook.html#etag-reference), specifically as implemented in #557, is for the `POST` request method for which, according the both the above RFCs, webhook endpoints should respond with a `412 (Precondition Failed)` status code.

Even since the refactoring of the webhook code, Metacontroller only accepts responses with a `304` status.

Some web frameworks that provide automated `ETag` support do so with stricter compliance to the RFC and thus break Metacontroller since `412` is not a supported status code (for example, the [Sinatra framework](https://github.com/sinatra/sinatra/blob/4f5ca85d408b2a64d9a0245eb5a1a14eaa2366b2/lib/sinatra/base.rb#L632) for Ruby, likely others).

This seems fairly straightforward to fix by extending the conditional logic in Metacontroller to also check for a `412`.

jgnagy:
Fixes #850

Adds support for `412 (Precondition Failed)` responses from controllers when the provided `If-None-Match` header contains the `ETag`. As described in #850, this is the expected behavior of HTTP/1.1 as described in [RFC7232](https://www.rfc-editor.org/rfc/rfc7232#section-3.2).

I also took the liberty of updating the docs.

### DIFF
--- a/docs/src/api/hook.md
+++ b/docs/src/api/hook.md
@@ -43,11 +43,11 @@ Within a `webhook`, the `service` field has the following subfields:
 
 ### Etag Reference
 
-More details in [rfc2616](https://datatracker.ietf.org/doc/html/rfc2616).
+More details in [rfc7232](https://www.rfc-editor.org/rfc/rfc7232).
 
 Etag is a hash of response content, controller that supports etag notion should add "ETag" header to each 200 response.
-Metacontroller that supports "ETag" should send "If-None-Match" tag with value of ETag of cached content.
-If content have not changed controller replies with "304 Not modified", otherwise it sends 200 with "ETag" header.
+Metacontrollers that support "ETag" should send the "If-None-Match" header with value of ETag of cached content.
+If content has not changed, controller should reply with "304 Not modified" or "412 Precondition Failed", otherwise it sends 200 with "ETag" header.
 
 This logic helps save traffic and CPU time on webhook processing.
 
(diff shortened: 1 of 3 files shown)


docs/src/api/hook.md
```diff
@@ -46,8 +46,8 @@ Within a `webhook`, the `service` field has the following subfields:
 More details in [rfc2616](https://datatracker.ietf.org/doc/html/rfc2616).
```
grzesuav, line 46: @jgnagy can you update the reference here ?

grzesuav (changes requested):

grzesuav:
thanks @jgnagy , added small comment. Thanks for the PR, approved the tests

grzesuav (approved):

Issue #731, opened 2022-12-23
Events from related resources not being received after a while


rolandkool:
When using the customize hook to read additional resources, I've noticed that after a while (not sure how long, I suspect about an hour) changes on those related resources are no longer received by the controller.

The customize hook is also called every hour, which does then pick up the changes that were made on the related resources.

I assume this is a bug as for the first hour or so it all works just fine.

grzesuav:
Can you elaborate more on the configuration you have ? 
1. Definition of controller CRD
2. Does the spec of related resource changes ?

Can you simulate the behavior, i.e. edit the related resource and see in metacontroller logs why it decides to not do anything ?

rolandkool:
The controller definition does not contain any unusual configuration. Just a sync and customize hook like the examples.

What do you mean with 2 exactly? The customize hook code does return the same resource types (i.e. a configmap), but it may not always return the same resource instances. These can vary based on existence of labels or how they're named.

The controller does not log anything, unfortunately. At least not with default logging settings. On my own app, I just don't see any calls coming in when there should be.

rolandkool:
I've done some more testing and have a way for you to reproduce. The issue can be easily reproduced using the [Secret Propagation Example](https://github.com/metacontroller/metacontroller/tree/master/examples/secretpropagation)

Deploy the controller and validate it works by making a change to either the source Secret or by adding/removing the `propagate` label on a namespace. Now wait a little over 10 minutes (11 to be on the safe side) and try again. Changing the source Secret or adding/removing a label on a namespace will not trigger the controller anymore (until the controller gets restarted).

If you keep making changes to resources that are watched by the customize hook within the 10 minutes interval, the changes are picked up immediately. The customize hook keeps working.

Could it be some kind of connection timeout / interruption that does not get refreshed / restored?

I've tested this only with GKE, but it happens both with the controller running as a pod on the cluster as well as from my own laptop connecting to the Kubernetes API.

grzesuav:
thanks, on local laptop you use `kind`/`minikube` ? What version of kubernetes ?

rolandkool:
On local laptop I don't use `kind`, but metacontroller connects to API of GKE using my own credentials.

I've tested against GKE 1.22 and 1.23.

rolandkool:
[...]

PR #886, opened 2023-10-17
fix(related_resources): Call getCustomizeHookResponse instead of getCachedCustomizeHookResponse

newtondev:
The customize hook cache would reach TTL and expire; the cached item would no longer be found. The loop in find findRelatedParents would trigger the continue which would skip over all the related rules to be processed.

By replacing getCachedCustomizeHookResponse with getCustomizeHookResponse, if the cached item is not found, it would then trigger the customize hook call and update the cache. The related items would then continue to be processed.

fixes: https://github.com/metacontroller/metacontroller/issues/731


grzesuav:
nice catch !

grzesuav (approved):

Issue #986, opened 2024-05-07
Feature request: Disable webhook calls for changes in status of parent CR


Shashankft9:
We have some CRDs built on top of cluster API that keep on getting updated by their controllers every minute, one such example is a cluster CRD, which we keep on updating every minute to show the health in status. Now using metacontroller, we run some webhooks through decorator where this CRD is the parent, and because its status keeps on getting updated every minute, the webhook is also called along with it, eventually causing too much unnecessary traffic in between.

This could be avoided if somehow we can block metacontroller to call the webhook on status change, but this should not be the default behavior, it should be opted for, maybe through the decorator CR? 
More context here: https://kubernetes.slack.com/archives/CA0SUPUDP/p1715069539278159

If this makes sense and if it is to be considered for a feature, I can help with it!

cc @grzesuav 

- [x] enable this feature at parent resources for composite controller and decorator controller
- [ ] fix update enqueue for parent resources that are being deleted if this feature is enabled
- [ ] enable this feature for attachement/child resources

grzesuav:
To summarize - it is decorator controller and one of the `resources` (https://metacontroller.github.io/metacontroller/api/decoratorcontroller.html#resources) have frequent status update ? 


Some ideas to consider:
1. Extend api,i.e.
```yaml
apiVersion: metacontroller.k8s.io/v1alpha1
kind: DecoratorController
metadata:
  name: service-per-pod
spec:
  resources:
  - apiVersion: apps/v1
    resource: statefulsets
    **ignoreStatusUpdate: true**
  attachments:
  - apiVersion: v1
    resource: services
  hooks:
    sync:
      webhook:
        url: http://service-per-pod.metacontroller/sync-service-per-pod
        timeout: 10s

```
(actually would make sense to add it also to `attachements` and composite controller also), however it would need to be filtered **before** sending `sync`/`finalize` request - need to think if it is doable

2.Add more generic concept similar to https://metacontroller.github.io/metacontroller/api/controllerrevision.html#parent-patch, so expose a field pointing to part of the object controller is interested about - however it can be not easy to implement - the previous value of the part of object under that path would need to be cached somewhere. Need to think a bit on this.

```yaml
apiVersion: metacontroller.k8s.io/v1alpha1
kind: DecoratorController
metadata:
  name: service-per-pod
spec:
  resources:
  - apiVersion: apps/v1
    resource: statefulsets
    **path: spec**
  attachments:
  - apiVersion: v1
    resource: services
[...]

PR #1000, opened 2024-06-10
feat: #986 - Ignore status changes for parent resources in dec and cc controller

Shashankft9:
fixes part of #986

- introduces a field `ignoreStatusChanges` in the API of both - decorator and composite controller
- if its set to `true`, only spec changes or label/annotation changes will be honored for reconcilation.

### DIFF
--- a/pkg/apis/metacontroller/v1alpha1/types.go
+++ b/pkg/apis/metacontroller/v1alpha1/types.go
@@ -58,9 +58,10 @@ type ResourceRule struct {
 }
 
 type CompositeControllerParentResourceRule struct {
-	ResourceRule    `json:",inline"`
-	RevisionHistory *CompositeControllerRevisionHistory `json:"revisionHistory,omitempty"`
-	LabelSelector   *metav1.LabelSelector               `json:"labelSelector,omitempty"`
+	ResourceRule        `json:",inline"`
+	RevisionHistory     *CompositeControllerRevisionHistory `json:"revisionHistory,omitempty"`
+	LabelSelector       *metav1.LabelSelector               `json:"labelSelector,omitempty"`
+	IgnoreStatusChanges *bool                               `json:"ignoreStatusChanges,omitempty"`
 }
 
 type CompositeControllerRevisionHistory struct {
@@ -224,9 +225,10 @@ type DecoratorControllerSpec struct {
 }
 
 type DecoratorControllerResourceRule struct {
-	ResourceRule       `json:",inline"`
-	LabelSelector      *metav1.LabelSelector `json:"labelSelector,omitempty"`
-	AnnotationSelector *AnnotationSelector   `json:"annotationSelector,omitempty"`
+	ResourceRule        `json:",inline"`
+	LabelSelector       *metav1.LabelSelector `json:"labelSelector,omitempty"`
+	AnnotationSelector  *AnnotationSelector   `json:"annotationSelector,omitempty"`
+	IgnoreStatusChanges *bool                 `json:"ignoreStatusChanges,omitempty"`
 }
 
 type AnnotationSelector struct {
--- a/pkg/controller/composite/controller.go
+++ b/pkg/controller/composite/controller.go
@@ -23,6 +23,7 @@ import (
 	commonv2 "metacontroller/pkg/controller/common/api/v2"
 	"metacontroller/pkg/hooks"
 	"metacontroller/pkg/logging"
+	"reflect"
 	"sync"
 	"time"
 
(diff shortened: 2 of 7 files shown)


pkg/apis/metacontroller/v1alpha1/types.go
```diff
@@ -58,9 +58,10 @@ type ResourceRule struct {
 }
 
 type CompositeControllerParentResourceRule struct {
-	ResourceRule    `json:",inline"`
-	RevisionHistory *CompositeControllerRevisionHistory `json:"revisionHistory,omitempty"`
-	LabelSelector   *metav1.LabelSelector               `json:"labelSelector,omitempty"`
+	ResourceRule        `json:",inline"`
+	RevisionHistory     *CompositeControllerRevisionHistory `json:"revisionHistory,omitempty"`
+	LabelSelector       *metav1.LabelSelector               `json:"labelSelector,omitempty"`
+	IgnoreStatusChanges bool                                `json:"ignoreStatusChanges,omitempty"`
```
grzesuav, line 64: it need to be a bool pointer, otherwise it will be always set

pkg/controller/composite/controller.go
```diff
@@ -169,21 +171,26 @@ func newParentController(
 			return nil, err
 		}
 	}
+	parentIgnoreStatusChanges, err := newCompositeIgnoreStatusChangesMap(resources, cc)
```
grzesuav, line 174: do we need this ? I believe it can be took from `cc` ?

Shashankft9: probably not needed for this PR, since in cc there is only one parent, but this could be useful when I create a subsequent PR for children resources, wdyt?

pkg/controller/composite/controller.go
```diff
@@ -359,6 +366,20 @@ func (pc *parentController) updateParentObject(old, cur interface{}) {
 	// different status (e.g. you have some incrementing counter).
 	// Doing that is an anti-pattern anyway because status generation should be
 	// idempotent if nothing meaningful has actually changed in the system.
+	if parentOld, ok := old.(*unstructured.Unstructured); ok {
```
grzesuav, line 369: 1. flip the condition, first check if  `ignoreStatusChange` is enabled, then do the comparison

pkg/controller/composite/controller.go
```diff
@@ -359,6 +366,20 @@ func (pc *parentController) updateParentObject(old, cur interface{}) {
...
+		if parentCur, ok := cur.(*unstructured.Unstructured); ok {
+			// if ignoreStatusChanges is set to true in the decorator controller, a parent object should only be
+			// enqueues if either there is a change in the generation or if there is a change in its labels/annotations,
+			// otherwise it will be ignored.
+			if pc.parentIgnoreStatusChanges.Compares(parentCur) {
```
grzesuav, line 374: not needed, why not inline to access 
```
if pc.cc.Spec.ParentResource.IgnoreStatusChanges
```
directly ?

We have controller created for specific `CR` so it is either enabled or disabled (or not set which is also disabled) which can be checked here

pkg/controller/composite/controller.go
grzesuav: it is not needed, you can check it directly like https://github.com/metacontroller/metacontroller/blob/master/pkg/controller/composite/controller.go#L620

grzesuav (changes requested):
Please make `ignoreStatusChange` bool pointers and remove `parentIgnoreStatusChanges` field from `parentController`

pkg/controller/decorator/controller.go
```diff
@@ -349,6 +350,32 @@ func (c *decoratorController) enqueueParentObjectAfter(obj interface{}, delay ti
...
 func (c *decoratorController) updateParentObject(old, cur interface{}) {
 	// TODO(enisoc): Is there any way to avoid resyncing after our own updates?
+	if parentOld, ok := old.(*unstructured.Unstructured); ok {
+		parentKind := parentOld.GetKind()
+		parentAPIVersion := parentOld.GetAPIVersion()
+		for _, parent := range c.dc.Spec.Resources {
+			resource := c.resources.Get(parent.APIVersion, parent.Resource)
+			if resource == nil {
+				// it's not a resource we care about
+				return
```
grzesuav, line 360: it should be `continue`

pkg/controller/decorator/controller.go
```diff
@@ -349,6 +350,32 @@ func (c *decoratorController) enqueueParentObjectAfter(obj interface{}, delay ti
 
 func (c *decoratorController) updateParentObject(old, cur interface{}) {
 	// TODO(enisoc): Is there any way to avoid resyncing after our own updates?
+	if parentOld, ok := old.(*unstructured.Unstructured); ok {
+		parentKind := parentOld.GetKind()
+		parentAPIVersion := parentOld.GetAPIVersion()
+		for _, parent := range c.dc.Spec.Resources {
+			resource := c.resources.Get(parent.APIVersion, parent.Resource)
```
grzesuav, line 357:
```
resource := c.resources.Get(parent.APIVersion, parent.Resource)
```
you can do the API version from https://github.com/metacontroller/metacontroller/blob/master/pkg/apis/metacontroller/v1alpha1/types.go#L55 directly (conversion from GVK to GVR - or opposite - is needed but we do it in other places)

pkg/controller/decorator/controller.go
grzesuav: you can also add comment that this should not happen, as the resourceMap dicovers all the resources from API server

pkg/controller/decorator/controller.go
```diff
@@ -349,6 +350,33 @@ func (c *decoratorController) enqueueParentObjectAfter(obj interface{}, delay ti
...
+				if parent.IgnoreStatusChanges != nil && *parent.IgnoreStatusChanges {
+					if parentCur, ok := cur.(*unstructured.Unstructured); ok {
+						if parentOld.GetGeneration() == parentCur.GetGeneration() {
+							if reflect.DeepEqual(parentOld.GetLabels(), parentCur.GetLabels()) && reflect.DeepEqual(parentOld.GetAnnotations(), parentCur.GetAnnotations()) {
+								return
+							}
+						}
+					}
+
+				}
```
grzesuav, line 376: @Shashankft9 as a last thing, can you fix this golangci-lint warning ? Remove the newline

grzesuav (approved):

grzesuav:
@Shashankft9 can you also rebase and squash commits to have just one ?

grzesuav:
@Shashankft9 thanks !

PR #1004, opened 2024-06-25
fix: allow update enqueue for parent resources being deleted when ignoreStatusChanges is being set to true

Shashankft9:
fixes part of https://github.com/metacontroller/metacontroller/issues/986

we should also handle the case when `deletiontimestamp` is added to the parent object, metacontroller should enqueue it. Otherwise, whenever a parent object is deleted, because only timestamp is added in metadata, it will come to `updateParentObject()` method which will ignore it because there has been no change in the generation, labels or annotations.


grzesuav (approved):

grzesuav:
@Shashankft9 can you amend commit message to fit the convention:
fix: message 

?

Thank you

PR #1021, opened 2024-10-08
chore(docs): Update helm-install.md

jeongukjae:
updated example command using helm oci since it is not v-prefixed anymore.

### DIFF
--- a/docs/src/guide/helm-install.md
+++ b/docs/src/guide/helm-install.md
@@ -21,7 +21,7 @@ helm install metacontroller deploy/helm/metacontroller-helm-v*.tgz
 Charts are published as [packages on ghcr.io](https://github.com/metacontroller/metacontroller/pkgs/container/metacontroller-helm)
 
 You can pull them like:
-* `HELM_EXPERIMENTAL_OCI=1 helm pull oci://ghcr.io/metacontroller/metacontroller-helm --version=4.11.19`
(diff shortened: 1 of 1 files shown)


newtondev (approved):

jeongukjae:
oh I didn't check the dco check, i will update soon

docs/src/guide/helm-install.md
```diff
@@ -21,7 +21,7 @@ helm install metacontroller deploy/helm/metacontroller-helm-v*.tgz
 Charts are published as [packages on ghcr.io](https://github.com/metacontroller/metacontroller/pkgs/container/metacontroller-helm)
 
 You can pull them like:
-* `HELM_EXPERIMENTAL_OCI=1 helm pull oci://ghcr.io/metacontroller/metacontroller-helm --version=v2.2.5`
+* `HELM_EXPERIMENTAL_OCI=1 helm pull oci://ghcr.io/metacontroller/metacontroller-helm --version=4.11.19`
```
grzesuav, line 24: it is actually example, it would be better to set `--version=<version>` instead any fixed one

grzesuav (changes requested):

docs/src/guide/helm-install.md
jeongukjae: fixed!

newtondev (approved):

grzesuav (approved):

PR #1035, opened 2024-11-15
chore(examples): update python examples to use python 3.11

newtondev:
Updated python version as well as python example code to use python 3.11.

Updated nginx image to use version 1.26.2.

Converted python2 code to python3.

Added examples/kind-load-images.sh utility script that will load the container images used in the examples to a running kind cluster.


grzesuav:
thanks for this @newtondev

Issue #1037, opened 2024-11-20
Implement ServerSideDiff - for comparing existing and desired resources


grzesuav:
# Problem statement

Metacontroller logic of diffing existing vs desired is lacking a lot of subtle nuances  - like ordering of fields or comparing the units. 

We could use Server Side Diff as argo does - https://github.com/argoproj/gitops-engine/pull/522 and allow to use it 
- as default on a flag
- per parent CRD definition
- maybe on more granular case

Of course this feature could be potentially dangerous and bring some unexpected behavior, but I believe it is much better approach and more sustainable in long run.

List of bugs related to server side diff in argocd- https://github.com/argoproj/argo-cd/issues?q=is%3Aissue+ServerSide+Diff

grzesuav:
https://github.com/argoproj/argo-cd/pull/13663/files example implementation in argocd

Issue #743, opened 2023-01-15
Metacontroller improperly diffs observed and desided state


remeq:
Hi,

I have following obeserved and desired states:

Observed:

```{
    "apiVersion": "serving.knative.dev/v1",
    "kind": "Service",
    "metadata": {
        "annotations": {
            "metacontroller.k8s.io/last-applied-configuration": '{"apiVersion":"serving.knative.dev/v1","kind":"Service","metadata":{"labels":{"controller-uid":"35636d54-fa7b-4daf-87fc-acab9217516d","svcname":"onramp2-bmwgarage"},"name":"prmtt1","namespace":"metac"},"spec":{"template":{"metadata":{"annotations":{"autoscaling.knative.dev/max-scale":"1","autoscaling.knative.dev/min-scale":"1","autoscaling.knative.dev/target":"1"},"labels":{"app":"onramp2-bmwgarage","svcname":"onramp2-bmwgarage"}},"spec":{"containers":[{"image":"docker.io/onramp2/prm1metac:latest","ports":[{"containerPort":8080,"name":"http1"}]}]}}}}',
            "serving.knative.dev/creator": "system:serviceaccount:metacontroller:metacontroller",
            "serving.knative.dev/lastModifier": "system:serviceaccount:metacontroller:metacontroller",
        },
        "creationTimestamp": "2023-01-11T12:53:02Z",
        "generation": 1,
        "labels": {
            "controller-uid": "35636d54-fa7b-4daf-87fc-acab9217516d",
            "svcname": "onramp2-bmwgarage",
        },
        "managedFields": [
            {
                "apiVersion": "serving.knative.dev/v1",
                "fieldsType": "FieldsV1",
                "fieldsV1": {
                    "f:status": {
                        ".": {},
                        "f:conditions": {},
                        "f:latestCreatedRevisionName": {},
                        "f:observedGeneration": {},
                        "f:url": {},
                    }
                },
                "manager": "controller",
                "operation": "Update",
                "subresource": "status",
                "time": "2023-01-11T12:53:02Z",
            },
            {
                "apiVersion": "serving.knative.dev/v1",
                "fieldsType": "FieldsV1",
                "fieldsV1": {
                    "f:metadata": {
                        "f:annotations": {
                            ".": {},
                            "f:metacontroller.k8s.io/last-applied-configuration": {},
                        },
                        "f:labels": {".": {}, "f:controller-uid": {}, "f:svcname": {}},
                        "f:ownerReferences": {
[...]

Issue #501, opened 2022-04-09
[Feature proposal] Consider switching to server side apply (SSA)


grzesuav:
# Description

[Server side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) or SSA, was introduced in 1.16, reaching GA over time. It basically mimic `kubectl apply` behavior in more declarative way, managed by kubernetes API server. General benefit would be that `metacontroller` could drop its [dynamic apply](https://metacontroller.github.io/metacontroller/api/apply.html#dynamic-apply) implementation. However we would need to make sure that they are equivalent. As far I can tell, the list/map handling is [driven by markers on CRD schema](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy) so it would be actually better, as now metacontroller implementations tries to ["guess" when to merge lists](https://github.com/metacontroller/metacontroller/blob/master/pkg/dynamic/apply/apply.go#L209) and in SSA user declares it in CRD.

# TODO before implementation
1. check how use SSA from client-go / controller-runtime
2. verify compatibility of SSA versus metacontroller dynamic apply
3. document and check requirements for handling list merges
4. Check how it works with builtin types (i.e. does it considers all the lists from metacontroller "guessing" list also as candidates for merging)

# Challenges
1. Explaining users how they should write CRD schema in order to have predictable list merges. 
2. Conflicts handling etc

## Related issues
* https://github.com/metacontroller/metacontroller/issues/650
* https://github.com/metacontroller/metacontroller/issues/566
* https://github.com/metacontroller/metacontroller/issues/562
* https://github.com/metacontroller/metacontroller/issues/444

grateful345:
status:
  inputs:
    persistentvolumeclaims:
      total: 20
  ...
For each output resource, we will report the total number of objects owned by this parent across all map keys. In addition, we will automatically aggregate conditions found on output objects, and report how many objects we own with that condition set to True.

For example:


status:
  ...
  outputs:
    volumesnapshots:
      total: 100
      ready: 97
  ...

grzesuav:
Done by @woehrl01 , available in v4.12.0 via https://github.com/metacontroller/metacontroller/pull/1061

PR #1061, opened 2025-03-02
feat: support server side apply

woehrl01:
Adds Server Side Apply support. Additionally, it keeps track of the object generation to prevent excessive calls to the API server.

See #1037
See #501
See #743


newtondev:
Hi @woehrl01 thanks for working on this. From what I can see it LGTM. I will run it on my cluster with a few more tests just to be sure.

newtondev:
@grzesuav can you also please take a look at the code and see if you have any suggestions or concerns. Thanks

pkg/controller/common/manage_children.go
```diff
@@ -197,15 +207,98 @@ func deleteChildren(client *dynamicclientset.ResourceClient, parent *unstructure
...
+					_, err := client.Namespace(obj.GetNamespace()).Update(context.TODO(), oldObj, metav1.UpdateOptions{})
+					if err != nil {
+						logging.Logger.Error(err, "Failed to remove last applied annotation from observed object", "parent", parent, "child", obj)
+						errs = append(errs, err)
+						continue
+					}
+				}
+			}
+
+			patched, err := client.Namespace(obj.GetNamespace()).Patch(context.TODO(), obj.GetName(), types.ApplyPatchType, data, metav1.PatchOptions{
```
newtondev, line 276: I like how you used PatchOptions instead of ApplyOptions and making it SSA by specifying types.ApplyPatchType.

pkg/controller/common/manage_children.go
```diff
@@ -197,15 +207,98 @@ func deleteChildren(client *dynamicclientset.ResourceClient, parent *unstructure
...
+				FieldManager: "metacontroller",
```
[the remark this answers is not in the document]
woehrl01: Makes sense, I justed added this as a cli option

pkg/controller/common/manage_children.go
woehrl01: Thank you!

pkg/controller/common/manage_children.go
```diff
@@ -42,6 +46,10 @@ func ApplyUpdate(orig, update *unstructured.Unstructured) (*unstructured.Unstruc
 	if err != nil {
 		return nil, err
 	}
+
+	//prevent setting last applied values in the new object
+	nullifyLastAppliedAnnotation(update)
```
grzesuav, line 51: should it happen only when SSA apply strategy is used, right ?

grzesuav:
hi @woehrl01 and thanks for PR !

I am AFK this weekend but will test locally on Monday

My initial thought :
* maybe instead of flag `--useServerSideApply` we could add `--apply-strategy` with default option `dynamic-apply` (which would refer to https://metacontroller.github.io/metacontroller/api/apply.html#dynamic-apply ) and another strategy (`server-side-apply`/`ssa`) which would be this one, maybe marked as experimental 
    * `--ssa-field-manager` -> `--apply-strategy-ssa-field-manager` maybe ? 
* the documentation https://metacontroller.github.io/metacontroller/api/apply.html would need to be extended to cover both options
* https://metacontroller.github.io/metacontroller/guide/configuration.html also should cover new flag
* regarding hashing approach - how does it behave after metacontroller restart ? I guess the cache will be created from scratch, right ? Not saying this is wrong, just want to understand implications
    * I have another issue open https://github.com/metacontroller/metacontroller/issues/1037 for implement dry-run ssa diff as alternative for diffing for dynamic apply which could be also used in future here as alternative strategy
    


let me know what do you think @woehrl01 @newtondev

pkg/controller/common/manage_children.go
woehrl01: actually no. This line fixes a bug where if you simply return the "observed" as a "expected" child, that you recursively end up with the whole object in "last applied" annotation, filling up ETCD (you don't want to know why I figured that out 😅 )

woehrl01:
Thank you regarding your points!

- I'm fine changing to general `--apply-strategy` seams like a reasonable generalisation. 
- Documentation update makes sense
- Regarding the hashing approach: Yes it implies that after a restart its filling up the cache and everything is applied again, which should be a noop if no change. I believe this is fine, at least in our cluster where we manage about 200k of objects we haven't seen an issue. Do you have any particular concern?

woehrl01:
Regarding your last comment for server side diff. I'm not sure that it makes sense in the context of metacontroller. For argocd it's nice, because you can preview the changes and manually apply the diff. For metacontroller this is not a valid use case as it will simply override "everything it get". Using serversideapply in my opinion is the only reasonable alternative to the existing approach. As I don't believe that that server side diff creates less pressure on the API server compared to SSA. 

I'm open for any opinions here, as I could likely miss something obvious.

grzesuav:
> Regarding your last comment for server side diff. I'm not sure that it makes sense in the context of metacontroller. For argocd it's nice, because you can preview the changes and manually apply the diff. For metacontroller this is not a valid use case as it will simply override "everything it get". Using serversideapply in my opinion is the only reasonable alternative to the existing approach. As I don't believe that that server side diff creates less pressure on the API server compared to SSA.
> 
> I'm open for any opinions here, as I could likely miss something obvious.

This part would be only to determine if there are any changes to apply, even not using SSA to finally apply the changes, I do not consider it tied with actually chosen apply method. It could also prevent some issues when using metacontroller dynamic apply. And yes, the API server usage will be most likely the same.

pkg/controller/common/manage_children.go
grzesuav: hmm got it, will try to recreate the issue

grzesuav:
Let me actually think how e2e tests can be adjusted to cover both cases 🤔

grzesuav:
Also, there is no need to change anything in your PR regarding the diff calculation, if I implement it it will be independent, and on different flag

woehrl01:
@grzesuav Just updated the changes as discussion. I hope you don't mind reformatting the docs, I have "prettier" auto-formatting enabled for my markdown docs.

docs/src/api/apply.md
```diff
@@ -260,13 +253,77 @@ If any of these are blockers for you,
...
+
+### Enabling Server-Side Apply
+
+To enable SSA in Metacontroller, configure the controller with:
+
+```sh
+--apply-strategy=server-side-apply
+```
+
+This setting ensures Metacontroller applies resources using Kubernetes-native SSA rather than dynamic apply.
```
grzesuav, line 323: maybe add a link to https://kubernetes.io/docs/reference/using-api/server-side-apply/ at the end ?

docs/src/api/apply.md
```diff
@@ -260,13 +253,77 @@ If any of these are blockers for you,
...
-  * A list of scalars (not objects) that should be merged as if the
+- Unconventional associative list keys
+  - An associative list that doesn't use one of the conventional keys.
+- Multi-field associative list keys
+  - A key that's composed of two or more fields (e.g. both `port` and `protocol`).
+- Scalar-valued associative lists
+  - A list of scalars (not objects) that should be merged as if the
     scalar values were field names in an object.
+
+## Server-Side Apply
```
grzesuav, line 267: can you mention this is new feature, so advise cautions when using it ?

grzesuav (dismissed):
Thanks for this PR, I will do some testing later this week, and if everything will be ok I will merge it, which will trigger release

grzesuav:
I simplified the e2e tests so we are focusing now on last 4 releases of kubernetes

grzesuav:
There are few minor linting errors

grzesuav:
Thanks ! Will try to adjust test to be possible to specify different apply method arguments, so we could essentially have both flows covered

woehrl01:
Great! Please let me know if I should do additional adjustments.

grzesuav:
it took me a while as github is not as flexible as gitlab - https://github.com/metacontroller/metacontroller/pull/1079 however @woehrl01 can you:
* make a copy of `manifest/dev` called `manifests/dev-ssa` with additional argument to run using server side apply
* add to https://github.com/metacontroller/metacontroller/blob/master/.github/config/kind-matrix-config.json version of each config with `dev-ssa` set as variant ?

it will make e2e run with those option, which would be ultimate test


Thanks !

woehrl01:
@grzesuav Absolutely! Done.

woehrl01:
@grzesuav saw some actions still failing, hope the last push fixes the validation around docs and linting

grzesuav:
there are some issues related with patching and missing field in schema, will take a look in following days

grzesuav:
`daemonjob/test.sh...FAILED`

woehrl01:
Do you have more output? I can't see the message of the error, it cuts at the parent object level.

I now push a doc change to fix that last linting error

woehrl01:
@grzesuav found the issue and push a fix, it seems that `terminationGracePeriodSeconds` just hat the wrong indentation

grzesuav:
@woehrl01 as a final step, can you rebase and squeeze all commit to just one ? It makes it more readable for viewing git history.

Thanks for all your works !

woehrl01:
@grzesuav You're welcome! I just pushed the squash (after seeing that the tests are now passing)

grzesuav:
https://github.com/metacontroller/metacontroller/pull/1081

woehrl01:
Will do later today, should be ready latest tomorrow.

grzesuav (approved):

PR #1121, opened 2025-07-11
fix: handle updatestrategy correctly for SSA

woehrl01:


grzesuav:
@woehrl01 I run this locally by:
* `make image`
* `kind load docker-image localhost/metacontroller:dev`
* ` kubectl apply -k manifests/dev-ssa`

from your branch.

Then I tried to run tests, the first one it is failing one is `examples/catset/` - I ran `./test.sh` - it seems that after applying manifests it tries to create and delete child pods in a loop. Haven't got more time to look into it but have it on my mind. Are you able to recreate this on our end ? It might allow you to narrow down the issue locally

grzesuav:
Looking at the looks it seems like the child pod which is creates is instantly orphaned and adopted :thinking:

woehrl01:
@grzesuav thank you that's a good starting point. I'll have a look

woehrl01:
@grzesuav I think i good it (haven't tested localy), but there was a diff, that with SSA I was deleting and creating the object in the same loop

pkg/controller/common/manage_children.go
```diff
@@ -232,184 +233,301 @@ func lastUpdateCacheKey(client *dynamicclientset.ResourceClient, obj *unstructur
...
+
+	cacheLock.Lock()
+	lastUpdatedCache[lastUpdateCacheName] = &lastUpdate{
+		hash:               hash,
+		resourcegeneration: patched.GetGeneration(),
+	}
+
+	logging.Logger.Info("Cache updated", "name", lastUpdateCacheName)
+
+	cacheLock.Unlock()
```
grzesuav: can we move it just after lock acquiring with defer keyword ?

PR #1189, opened 2026-03-23
fix(helm): deploy as Deployment

yajo:
Metacontroller is stateless, so no need to deploy as a StatefulSet.

This way, the scheduler has more flexibility.

@moduon MT-14138


grzesuav:
@yajo please follow insructions  from DCO first. Did you checked there is no change needed in e2e examples or kustomize after changing that ?

yajo:
Let's see now.

PR #1211, opened 2026-07-03
fix(endpoints Config): Add min tls 1.2, few after review fixes

grzesuav:
Few adjustments after introducing TLS

cc: @flavio


flavio (approved):
LGTM

PR #1231, opened 2026-07-09
fix(hooks): log invalid webhook configuration instead of failing silently

AruneshDwivedi:
When a hook specifies neither a full url nor both service and path, NewWebhookExecutor returned the error without logging it, so a typo'd or missing webhook url produced no operator-visible signal. The error is now logged with the controller and hook type so the misconfiguration is surfaced.


grzesuav (approved):

grzesuav:
@AruneshDwivedi thanks for the pr !

